PR #34, opened 2016-05-30
New public method LayerGroup.pm.layerChanged()

### LINKED ISSUE #33: pm.toggleEdit does not work after new layer contents is added
When new layer contents is added after the `pm` plugin was initialized, it does not handle `pm:edit` events for newly added elements.

Like this:

```
var geoJsonLayer = L.geoJson().addTo(map);
geoJsonLayer.addData(json);
geoJsonLayer.addEventListener('click', function() {
    geoJsonLayer.pm.toggleEdit();
});
```

Clicking on the `geoJsonLayer` does not enter the edit mode

xmojmr:
This fixes #33. After new layer contents is added, calling `geoJsonLayer.pm.layerChanged()` resolves the issue

### DIFF
--- a/src/js/L.PM.LayerGroup.js
+++ b/src/js/L.PM.LayerGroup.js
@@ -3,7 +3,16 @@ L.PM.Edit.LayerGroup = L.Class.extend({
     initialize: function(layerGroup) {
         var self = this;
         this._layerGroup = layerGroup;
-        this._layers = layerGroup.getLayers();
+        this.layerChanged();
+    },
+    layerChanged: function() {
+        if (this._layers) {
+          for(var i=0; i<this._layers.length; i++) {
+              this._layers[i].off('pm:edit', this._fireEdit);
+          }
+        }
+            
+        this._layers = this._layerGroup.getLayers();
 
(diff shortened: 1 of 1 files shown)


codeofsumit:
@xmojmr thank you very much for you pull request.

I really do appreciate you taking the time to fix #33, but in your solution the user (aka you) has to manually call `layerChanged()` whenever data is added to a geoJsonLayer via `addData()`.
I wanted a solution that just works whenever `addData()` is called.

So I asked for help [here](https://github.com/Leaflet/Leaflet/issues/4609#issuecomment-222506238) in the leaflet repo and they gave me the clue I needed to fix this.

I just created a new release that should fix this issue. Please let me know if it works or not over here: #33.
https://github.com/codeofsumit/leaflet.pm/releases/tag/v0.6.2

If you're using bower, you should be able to update the library with `bower install leaflet.pm --save`

If you're interested how I did it, here is the Pull Request:
https://github.com/codeofsumit/leaflet.pm/pull/36/files

xmojmr:
Your solution works for me. Thank you

Issue #14, opened 2016-02-10
Icons missing


codeofsumit:
Not sure how bower can handle images (icons). Got to do something about it or make it a requirement that devs provide icon paths.

codeofsumit:
will be fixed with the next release

PR #40, opened 2016-08-05
Change from iconUrl to className

kmpm:
Use css (which is relative to css) for
images. This will stop icons from being missing
and won't require a absolute path since everything
is in a relative path to the css file.

This should fix the issue #14.


codeofsumit:
love it, thank you and sorry for the delay

PR #71, opened 2016-09-07
Snapping Feature

codeofsumit:


codeofsumit:
Test a little more, do some cleanup but yeah, it's working already.
Merge after cleanups and some docs.

PR #88, opened 2016-10-03
Now makes sure .pm is avaible before invoking.

thomaspersson:
Was issues with other Leaflet plugins, where leaflet.pm interfered, might be a better more isolated solution but this was a quick fix to make sure no JS errors was thrown.

### DIFF
--- a/src/js/Draw/L.PM.Draw.Line.js
+++ b/src/js/Draw/L.PM.Draw.Line.js
@@ -37,8 +37,10 @@ L.PM.Draw.Line = L.PM.Draw.extend({
         // fire drawstart event
         this._map.fire('pm:drawstart', { shape: this._shape });
 
-        // toggle the draw button of the Toolbar in case drawing mode got enabled without the button
-        this._map.pm.Toolbar.toggleButton(this.toolbarButtonName, true);
+        if (this._map.pm) {
+            // toggle the draw button of the Toolbar in case drawing mode got enabled without the button
+            this._map.pm.Toolbar.toggleButton(this.toolbarButtonName, true);
+        }
     },
     disable() {
         // disable draw mode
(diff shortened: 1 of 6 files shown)


codeofsumit:
Do you mind sharing a bit about the issue?

1. What other leaflet plugin did you use?
2. Is there a JSFiddle or something?

I might fix this issue more globally instead of putting ifs everywhere. But I need to know some details of your usecase. Normally, leaflet.pm would put the `PM` instance on all leaflet maps, polygons, polylines but not on markers and so on.

thomaspersson:
Yes, of course.

1. https://github.com/Leaflet/Leaflet.markercluster#readme'
2. No, I don't have a JS fiddle.

The scenario is the following. I got two seperate modules that does not share code between each other or through some other file. The only thing in common is import L from 'leaflet'.

The first module uses Leaflet and leaflet.pm and the other module uses Leaflet and leaflet.markercluster.

If I only use one of each module they both work fine. But if I load the same modules in a view I get an error from my module with leaflet.markercluster. The error is as follow:

leaflet.pm.min.js:1 Uncaught (in promise) TypeError: Cannot set property '_layerGroup' of undefined

If I remove the leaflet.pm import this error does not happened and if I use the safe if calls it does not appear and both modules work.

L.PM is avaible in the module that don't make use of leaflet.pm but it's the instances such as layers, map etc that don't have .pm avaible.

codeofsumit:
ok from what I can see, this isn't an issue with leaflet.pm and markercluster working together (I can use them just fine). More that you're calling leaflet.pm functions on layers where leaflet.pm isn't initiated.

Could it be that you accidentally put some events on your layers on the map of markercluster, use leaflet.pm functionality in their callbacks (like `enable()` on click) and then the error gets thrown?

If you use leaflet.pm functions, those functions shouldn't have to check if leaflet.pm is available.

I think you have two options:

1. check your code if you're accidentally calling leaflet.pm functions on the map where you DON'T have leaflet.pm imported and fix that.
2. Import leaflet.pm in the other module as well.

Do you think this might be the issue? If not, I'd need more code from you to reproduce.

codeofsumit:
to be honest I don't get how you can get into the leaflet.pm code in the first place, when e.g. `polygon.pm` isn't available. So how could an error inside `enable()` be thrown? I think I need more code.

codeofsumit:
Please try this branch and see if it works. If yes, I'll release the fix tomorrow.
https://github.com/codeofsumit/leaflet.pm/tree/plugin-compatibility

PR #132, opened 2017-02-08
snap for circlemarker

alanblins:
Snap doesn't work when we use L.circleMaker just for L.marker.


codeofsumit:
hey, thanks for the PR. Didn't even know about circleMarkers 😅. I'll check it out and merge/implement asap.

PR #152, opened 2017-04-25
Marker style options are now applied to the hint marker

SwankSwashbucklers:
Needed in my project because we want a custom marker icon for the hint marker.


codeofsumit:
thanks for the PR 👍

PR #154, opened 2017-05-04
Update bug with library font-awesome

lclpedro:


codeofsumit:
Hi thanks for the PR. As far as I can see the PR is missing the name changes for the classes in [this file](https://github.com/codeofsumit/leaflet.pm/blob/develop/src/js/Toolbar/L.PM.Toolbar.js).

Also please prefix the classes with `leaflet-pm`. So `icon-polygon` becomes `leaflet-pm-icon-polygon`.

👍 good catch, will get merged when the changes are made.

lclpedro:
Hi Friend. Added update leaflet-pm-icon-* 

Archives changed:
    - src/js/Toolbar/L.PM.Toolbar.js
    - dist/leaflet.pm.css
    - dist/leaflet.pm.min.js

PR #156, opened 2017-05-07
Webpack build

codeofsumit:


codeofsumit:
- [ ] Before merging, check the install via symlink in another web app to verify the `prepare` script is working

PR #178, opened 2017-07-19
More polyfills for IE

thomasvoigt:
Had some time to spare with IE11+ and found some more ES6 hickups (sorry for missing them the first time).

* More polyfills for IE:
  - Object.assign() is needed for snapping
  - Element.remove() is needed when removing control buttons
* demo.js dumbed down for IE11:
  - IE doesn't understand thoose fancy lambdy style expressions
  - object literals always need a key

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -2,11 +2,11 @@
 
 const map2 = L.map('example2').setView([51.505, -0.09], 13);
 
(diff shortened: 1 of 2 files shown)


codeofsumit:
Looks good, I'll have it tested by tomorrow and merge it 👍

PR #195, opened 2017-08-11
Don't use cheap-eval-source-map in main build

nparley:
This PM changes the build source map and minimise the source. `cheap-eval-source-map` is not production safe and should therefore only be in the development build. See https://webpack.js.org/configuration/devtool/.


codeofsumit:
hey, thanks for the PR! I'll try this out and merge asap!

codeofsumit:
looks good! Thanks for your contribution!

PR #198, opened 2017-08-16
Hole support

### LINKED ISSUE #150: Is it possible to draw a hole in polygon?
I want to draw a hole in polygon. For example, when user hold option key and draw new polygon inside existing polygon, it becomes a hole. 
something like this http://jsfiddle.net/erictheise/VaGy5/

codeofsumit:
- [x] Editing holes
- [x] Drawing holes

fixes #150


codeofsumit:
Editing Holes is finished.
Will work in drawing them now

codeofsumit:
drawholes

codeofsumit:
I guess this is mergeable 🤔

PR #203, opened 2017-08-25
Map drag fix

### LINKED ISSUE #202: If a polygon is dragged such that the cursor leaves the map container, map is no longer scrollable
1. Create a Polygon

screen shot 2017-08-25 at 3 37 38 pm

2. Drag the Polygon such the the cursor leaves the map container.

screen shot 2017-08-25 at 3 37 44 pm

3. Click back into the map container, and find the Polygon still in the middle of a drag event.

4. After placing the polygon by clicking, the map is no longer draggable.

I believe this is caused by **_dragMixinOnMouseMove()** in *Drag.js*, which temporarily stores the state of the map's draggable status on mouse down. 

**_dragMixinOnMouseUp()** restores the map's dragging status, but isn't called if the mouse up event occurs outside the map container.

Thus, when you click back into the map container in **Step 2**, the temporary variable storing the map's draggable status is set to false (as the map isn't draggable at that moment).

I was able to fix it by moving 

> this._originalMapDragState = this._layer._map.dragging._enabled;

to **_initDraggableLayer()**, but I understand if that's not satisfactory (as it fails to take into account the user dynamically toggling the map's draggable status)

Mesonyx:
Closes #202 

If you drag a Polygon such that the cursor goes off the map, the **mouseup** event that resets the map's draggable state never fires. This means that when we click back into the map and the **mousedown** event binding saves the map's current draggable state, it is always **false**.

The basic idea of this PR is that we first record the map's draggable state upon draggable layer initialization, and then use an additional flag to see if it's "safe" to cache the map's draggable state during all subsequent **mousedown** events.

It's always safe upon first initializing the draggable layer, as no **mousedown** event could have possibly occurred.

After that, it only becomes safe upon  capturing a **mouseup** event within the map container. This way, if you click outside then click back in, it doesn't save the new (always **false**) draggable state, instead defaulting to the state detected during the initial **mousedown** event that started the Polygon drag!

### DIFF
--- a/src/js/Mixins/Drag.js
+++ b/src/js/Mixins/Drag.js
@@ -96,8 +96,6 @@ const DragMixin = {
 
         this._layer.on('mouseup', this._dragMixinOnMouseUp, this);
 
-        console.log(this)
-
         // listen to mousemove on map (instead of polygon),
         // otherwise fast mouse movements stop the drag
         this._layer._map.on('mousemove', this._dragMixinOnMouseMove, this);
--- a/src/js/Mixins/Drag.js
+++ b/src/js/Mixins/Drag.js
@@ -23,7 +23,7 @@ const DragMixin = {
             this._layer._map.dragging.enable();
         }
 
-        // if event fired, it's safe to cache the map draggable state on the next mouse down
+        // if mouseup event fired, it's safe to cache the map draggable state on the next mouse down
         this._safeToCacheDragState = true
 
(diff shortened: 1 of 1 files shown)


codeofsumit:
Hey, thanks for the PR - will check it this week! Sorry for the late reply I was on vacation.

codeofsumit:
nifty idea! Thanks for this fix and great first-time contribution!

PR #214, opened 2017-09-13
add option to pass event name to perform finish of polygon

themre:
as discussed here: https://github.com/codeofsumit/leaflet.pm/issues/147


src/js/Draw/L.PM.Draw.Line.js
```diff
@@ -48,9 +48,14 @@ Draw.Line = Draw.extend({
         this._map.on('click', this._createVertex, this);
 
         // finish on double click
-        if(this.options.finishOnDoubleClick) {
+        if (this.options.finishOnDoubleClick) {
             this._map.on('dblclick', this._finishShape, this);
         }
+        // finish on layer event 
+        // #http://leafletjs.com/reference-1.2.0.html#interactive-layer-click
+        if (this.options.finishOn) {
```
codeofsumit: Would be cool to add a default to the options object so `click` is used when no `finishOn` option is specified

demo/demo.js
```diff
@@ -143,7 +143,9 @@ map3.pm.addControls({
     drawMarker: true,
     drawPolygon: true,
     editPolygon: true,
-    drawPolyline: true,
+    drawPolyline: {
```
codeofsumit: the toolbar evaluation which buttons should be shown still works when there is an `Object` passed instead of `true` I assume?

codeofsumit: This is a cool way to pass options to the toolbar!
But I don't think they get applied right now - this would need work inside `Toolbar.js` to pass these options to the triggered functions (like Draw.enable).
Also would need documentation in the `README.MD`

themre: sorry, i left for vacations, but i'll look at these once I come back! thanks for the tips.

src/js/Draw/L.PM.Draw.Line.js
themre: would this be ok? `this._map.on(this.options.finishOn || 'dblclick', this._finishShape, this);`
I wouldn't put `click` since it would always clone polygon on first click :)

codeofsumit: the default should be inside the `options` object in the beginning of the class. It gets merged with the passed options, then you can just use `this.options.finishOn`.

demo/demo.js
themre: ok I looked a bit into this and perhaps it would be better to somehow refactor this whole code. because from what i'm seeing is that you have some global settings that apply to all actions (e.g.  snapping, dragging). also for edit button I noticed this:
 ```
afterClick: () => {
                this.map.pm.toggleGlobalEditMode({
                    snappable: true,
                    draggable: true,
                });
            },
```
.. which I would understand that it activates snapping and dragging for editing always. now, if we provided some object setting for individual tool ,we could fine-tune whether a tool needs snapping, dragging etc. or not. 
but since this involved much work, I would correct errors for now and maybe create this stuff in a separate PR.

src/js/Draw/L.PM.Draw.Line.js
themre: i see what you mean. but currently finishOnDoubleClick is turned off so I also added finishOn to null.

themre:
is there something maybe I missed? those stuff for adding Objects for options I would like to add in a separate PR.

codeofsumit:
sorry haven't had time to check it yet. Will do so over the weekend!

codeofsumit:
Sorry I haven't been able to look at it yet. Again: Will do so asap!

codeofsumit:
@themre thanks again for the PR. I reviewed it an added in the pieces missing (I'm still on it).
The new PR is here: https://github.com/codeofsumit/leaflet.pm/pull/233 - if you're interested you can check the commits what was missing.
Again: great idea!

I'll close this PR and finish the feature in #233. Great work man 👍

PR #219, opened 2017-09-20
Add mouseup event to line/polygon markers

newmanw:
I am looking to detect a vertex click or change for a line/polygon.  'mouseup' event will handle both.  Thoughts?

### DIFF
--- a/src/js/Edit/L.PM.Edit.Line.js
+++ b/src/js/Edit/L.PM.Edit.Line.js
@@ -142,6 +142,7 @@ Edit.Line = Edit.extend({
         marker.on('move', this._onMarkerDrag, this);
         marker.on('dragend', this._onMarkerDragEnd, this);
         marker.on('contextmenu', this._removeMarker, this);
+        marker.on('mouseup', this._onMouseUp, this);
 
         this._markerGroup.addLayer(marker);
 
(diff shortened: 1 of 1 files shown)


codeofsumit:
hey @newmanw - How do you think about a `pm:createVertex` event whenever there is a vertex added to the polygon?
If the payload includes the vertex and the current polygon shape - would it suit your needs as well?
Because to me this would fit more into the events of leaflet.pm (events for what's happening, not what the user is doing).

codeofsumit:
Also your PR adds this to the `Edit` part - but not in the `Draw` part? Is there a specific reason for that?

newmanw:
@codeofsumit currently there is no way that I know of that I can detect when a vertex is clicked.  I would like to display the lat/lng of the clicked/selected vertex.  

This is definitely an event based upon what the user is doing, however leaflet.pm creates the vertex markers, so there is no other way for my application to detect a click on a vertex.

Other thoughts on how this can be achieved?

codeofsumit:
@newmanw do you want the vertex while it's being DRAWN or do you want the vertex while an existing shape is EDITED? Or both?

newmanw:
@codeofsumit, thanks so much for the discussion!  

Currently I am only interested in EDIT as that is when existing vertices can be clicked.  That being said I could see value in a similar event for new DRAWN vertices as clients may want to display the lat/lng of those vertices as they are being drawn.

codeofsumit:
ah! Ok. Well I think this should be done for both, edit and draw.
An event that fires when a new vertex is created. Do you mind creating an issue with this as a request?
I'll implement this probably next week so I'm closing this PR now.

Thanks for the suggestion 👍

PR #220, opened 2017-09-20
Hint marker in map center rather than 0,0

newmanw:


codeofsumit:
Hey, thanks for the PR - what problem does this solve? Could you provide a short description?

newmanw:
@codeofsumit np.  I was having problems with initially seeing the hint marker when zoomed in.  If the hint marker is displayed at the map center, rather than 0,0 it will always been in the viewport.

PR #235, opened 2017-10-07
Add jsDelivr links

LukasDrgon:
I added [jsDelivr CDN links](https://www.jsdelivr.com/package/npm/leaflet.pm) to your readme as an alternative to unpkg. jsDelivr is the fastest opensource CDN available and built specifically for production usage. It can serve any project from npm with zero config just like unpkg, but offers a larger network and better reliability. We also have detailed usage stats for project maintainers.


codeofsumit:
Hi there - Thanks, I'll add other CDNs when users are requesting it.

PR #241, opened 2017-10-10
ADDED: `allowSelfIntersection` option (minor)

### LINKED ISSUE #239: New feature: prevent polygon self-intersection while editing
Hey

In our project we deal with simple polygons (shape consisting of straight, non-intersecting line segments)
and using Leaflet.Draw mostly because of `allowIntersection` feature.
You can see it in action on examples page https://leaflet.github.io/Leaflet.draw/docs/examples/full.html
1. Create a poly
2. Toggle edit mode
3. Drag a vertex to cross any segment - vertex will be deleted 

I want to migrate to leaflet.pm but without same feature it's absolutely impossible. Is there any chance to add such functionality or any workaround how to implement it using events?

Thanks for answer and amazing job!

codeofsumit:
fixes #239

### DIFF
--- a/src/js/Draw/L.PM.Draw.Cut.js
+++ b/src/js/Draw/L.PM.Draw.Cut.js
@@ -13,21 +13,16 @@ Draw.Cut = Draw.Poly.extend({
 
         // find all layers that intersect with `layer`, the just drawn cutting layer
         const layers = Object.keys(all)
-
-        // convert object to array
-        .map(l => all[l])
-
-        // only layers handled by leaflet.pm
-        .filter(l => l.pm)
-
-        // only polygons
-        .filter(l => l instanceof L.Polygon)
-
-        // exclude the drawn one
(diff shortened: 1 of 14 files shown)


src/js/Draw/L.PM.Draw.Cut.js
```diff
@@ -39,23 +41,32 @@ Draw.Cut = Draw.Poly.extend({
...
+                    const newL = L.geoJSON(g, l.options);
                     resultingLayers.push(newL);
                     newL.addTo(this._map);
+
+                    // give the new layer the original options
+                    newL.pm.enable(this.options);
+                    newL.pm.disable();
                 });
             } else {
+                console.log(l);
```
Luketep: Is that still required? ;)

codeofsumit: 🖕 fixed

PR #256, opened 2017-11-24
WIP: use rollup

themre:
I created a quick branch to use rollup. I added turf functions. it's still big because it's without uglify (have this problem https://github.com/Turfjs/turf/issues/1086), but it's size went from 1.3 to 0.95MB. also CSS images I copied next to CSS and didn't bundled them into CSS, because I didn't have time for embeding them.


themre:
well... @DenisCarriere did awesome job in converting turf modules to ES. I extracted some methods from jsts to separate utils file. final bundle is now almost 300kB smaller (final is 284 kB). webpack is not suitable for this, bundle is 634 kB big.

DenisCarriere:
👍 Sweet! Glad to hear it worked out, ideally we can keep making it smaller and smaller since there's a lot of dead code laying around a few dependencies, but it's a step forward!

codeofsumit:
@themre @DenisCarriere great to see progress in this but I still think these optimization should be upstream in the dependency tree. So either turf or jsts as I don't think leaflet.pm's built should optimize it's dependencies, but only it's own code.

@DenisCarriere is that something you will incorporate in the served turf bundle or would you suggest to put it into leaflet.pm (and essentially any other library that uses turf)?

codeofsumit:
guys, thanks for your work on this but I will close this now as the size optimizations should happen upstream in the dependency tree and leaflet.pm will only optimize it's own code.

PR #259, opened 2017-11-30
Cut events update

jpknapp:
when the cut event is fired it includes resulting layers from all the previous cut events, so if you cut several polygons each successive cut event accumulates previous resulting layers 

this update only includes resulting layers for the current cutted layer

### DIFF
--- a/src/js/Draw/L.PM.Draw.Cut.js
+++ b/src/js/Draw/L.PM.Draw.Cut.js
@@ -31,11 +31,11 @@ Draw.Cut = Draw.Poly.extend({
                 }
             });
 
-        // the resulting layers after the cut
-        const resultingLayers = [];
 
         // loop through all layers that intersect with the drawn (cutting) layer
         layers.forEach((l) => {
+            // the resulting layers after the cut
+            const resultingLayers = [];
             // find layer difference
             const diff = difference(l.toGeoJSON(), layer.toGeoJSON());
 


codeofsumit:
hey @jpknapp, thanks for the PR and taking the time to go ahead and tackle the problem yourself. Really appreciate it. However I was not able to reproduce the problem.
For me, cutting multiple polygons one after the other always returned the correct results.

Looking at the line of code you changed I'm also quite sure it'd stop serving all resulting polygons (e.g. if you cut a polygon in half, the result should be two polygons).

Can you please create a JS Fiddle that shows the problem? I'd love get to the bottom of this.

jpknapp:
http://jsfiddle.net/2odogvrr/
if you cut a single hole in the middle of 3 overlapping polygons, 3 cut events are fired with 1, 2, and 3 resultingLayers
https://i.imgur.com/HlOpr7w.png
The same thing happens if I use listeners on each individual layer.
When the cut event is fired for a layer, I only want the resulting layers from that layer (which will be 1 unless you cut the polygon in half).
I want to save the resulting layers to my database, but if multiple layers are cut in half resulting in 2 resulting layers (for each cutted layer) I cannot just loop through resultingLayers and save all those because resultingLayers includes everything from the previously fired events.
Does this make sense? 
sorry, I probably should have made an issue for this first...

codeofsumit:
@jpknapp thanks a ton for the explanation and your effort to solve it 👏 .
I understood now, was able to reproduce and can confirm your solution works 👍

Issue #249, opened 2017-11-13
Uncaught TypeError: Cannot read property 'lat' of null


galileopy:
I'm not sure if this error is due to leaflet.pm or is leaflet itself that has the bug, but certainly it didn't appear until I added leaflet.pm to my package.

The error appears when you add a polyline that has an array of arrays of [lat,lng], which leds me to believe it has something to do with the lack of support to MultiPolylines in leaflet.pm itself.

[Fiddle]( http://jsfiddle.net/t5s1yn3z/)

galileopy:
this  might be related, not sure though, https://github.com/Leaflet/Leaflet/issues/3280

galileopy:
leaflet.pm bug report, in case is weird edge case, https://github.com/codeofsumit/leaflet.pm/issues/249

codeofsumit:
yes, looks like that's exactly the reason. I'll investigate further and maybe I can add MultiPoly support within the next two weeks if there is demand now

perliedman:
👋 Hi everyone,

just to add to this, while checking https://github.com/Leaflet/Leaflet/issues/3280, it looked like at least the first problem is https://github.com/codeofsumit/leaflet.pm/blob/47336a37fa4d36a0a3137292b6b074c97321f0c8/src/js/Edit/L.PM.Edit.Line.js#L194, which assumes that things that are not polygons (which I assume are polylines) are not nested arrays, so it will assume each entry is a `LatLng`, which isn't true for MultLine.

If it makes you feel better, I actually discovered just now that Leaflet's `Polyline` actually support MultiPolyline as well...

codeofsumit:
thanks @perliedman. Yes I think with the support of holes, the groundwork for MultiPolylines and MultiPolygons is already in leaflet.pm. I specifically excluded them but apparently with some errors.
I'll sort it out 👍

codeofsumit:
just to give an update to this: obviously I wasn't able to make it work within 2 weeks. I'll get on it asap. I'm on a business trip right now and hope to get some time for this next week.

maleta:
I had this error when updating marker location with method `setLatLng` and my workaround for this was to remove existing marker and create new.

seanbeaton:
@codeofsumit, did you have any luck with this?

seanbeaton:
[...]

PR #279, opened 2018-03-01
(Very) basic multipolygon support

mpschaeuble:
I added some (very) basic multipolygon support to leaflet.pm. Don't know if its the right approach or whether multipolygons should be better handled as a LayerGroup of polygons, just thought it could be a starting point or help someone else.
At least it should solve issue #249.

### DIFF
--- a/src/js/Edit/L.PM.Edit.Line.js
+++ b/src/js/Edit/L.PM.Edit.Line.js
@@ -152,7 +152,7 @@ Edit.Line = Edit.extend({
 
     _initMarkers() {
         const map = this._map;
-        const coords = this._layer.getLatLngs();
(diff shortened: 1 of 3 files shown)


smeijer:
@codeofsumit could you take a look at this PR or #293 ?

codeofsumit:
@smeijer @mpschaeuble thanks for your work on this. If we introduce Multipolygon Support I would like to have it as general as possible with all possibilities covered.
I'm talking an infinite amount of polygons inside a multipolygon, each with an infinite amount of possible holes. These are quite the array structures to deal with.

Long story short: I'm working on it right now and I created a PR for you to take a look, comment, feedback and test if you're interested:
https://github.com/codeofsumit/leaflet.pm/pull/305/files

Issue #274, opened 2018-01-24
Drag-only edit mode


mpschaeuble:
For some use-cases it would be useful to allow dragging of existing features only in edit mode and disallow node editing.

PR #280, opened 2018-03-01
Add new edit option "disableNodeEditing"

mpschaeuble:
Fixes issue #274 
Adds a new option "disableNodeEditing".


codeofsumit:
@mpschaeuble thanks for this PR! I really appreciate it.
I went with another implementation though. If the option is set, I just create the markers without the draggable option. If you're interested in the implementation, check out this commit: https://github.com/codeofsumit/leaflet.pm/commit/dde74dce5365cf3cc7b0fde837d26c880f44d98b

Thanks again for your effort 👍

PR #293, opened 2018-05-03
fix(snapping): fix bug when the map contains a multi-geometry features (MultiPolygon / MultiLine)

### LINKED ISSUE #257: after firing map.pm.enableDraw(...) - on each "mousemove" event getting error "Cannot read property 'lat' of null"
Hey there,

First, I have to say that despite the error that promping on console, the behaviour of the functionality is great. The result is pure Rectangle and the working progress is great. 

**The operation :** 

I'm trying to draw few shapes using leaflet.pm, and I found some troubles while using the method:
`map.pm.enableDraw("Rectangle", {
      drawing: true,
      // snapping
      snappable: true,
      snapDistance: 20,
      allowSelfIntersection: true, 
      finishOn: 'dblclick', 
      templineStyle: {
        color: 'blue'
      },
      hintlineStyle: {
        color: 'blue',
        dashArray: [5, 5]
      },
      cursorMarker: true,
      pathOptions: {
        color: 'red',
        fillColor: 'orange',
        fillOpacity: 0.7
      },
      markerStyle: {
        opacity: 0.5,
        draggable: true,
      }
    });`

Then, when I'm moving with the cursor around the leaflet map I'm getting error.
I already debugged the leaflet module all the day and I found the source of the bug : 
 In the function : 
`
// @method latLngToLayerPoint(latlng: LatLng): Point

	// Given a geographical coordinate, returns the corresponding pixel coordinate

	// relative to the [origin pixel](#map-getpixelorigin)

latLngToLayerPoint: function (latlng) {
			     var projectedPoint = this.project(toLatLng(latlng))._round();
			     return projectedPoint._subtract(this.getPixelOrigin());
		}
`
The function getting the parameter _latlng_ that expecting to be type of _LatLng_ but instead he getting
an array of 17 elements of type _LatLng_.

arrayoflatlng

So - when executing the function `toLatLng(latlng)` it returns null, and then this null sended to 'this.project' method and then it throws an error to console. 

So I tried to change and manipulate the code in purpose to avoid error's on console, and I have done it, but the performance of the drawing Rectangle process (while moving the cursor around the map) became awful (just getting too slow when mouse move during the draw), so I can't really found a good solution to the error's on console.

* Needless to say that I already tried to wrap this code area by try & catch but I haven't find a solution that will work well.
* I also tried to send each element of array with for-loop, and also tried to send only the last/first point in the array. All this workarounds has led me to heavy performance issue (But the error's on console disappeared).

I want to say again - the performance when I getting the error's on console were great, but I want to avoid the error's that coming from the leaflet module.

By the way, I'm using the latest version of leaflet : "1.2.0+HEAD.1ac320b"

Thank you friends.

smeijer:
MultiPolygon layers trigger a bug in the snapping algorithm. This PR fixes this by breaking the MultiPolygon layers up before adding to the `_snapList`.

Fixes #257. More details explained in this comment: https://github.com/codeofsumit/leaflet.pm/issues/257#issuecomment-386079468

### DIFF
--- a/src/js/Mixins/Snapping.js
+++ b/src/js/Mixins/Snapping.js
@@ -195,9 +195,6 @@ const SnapMixin = {
                 else {
                     layers.push(layer);
                 }
-
-                map.off('pm:remove', this._handleSnapLayerRemoval, this);
-                map.on('pm:remove', this._handleSnapLayerRemoval, this);
             }
         });
 
@@ -224,6 +221,8 @@ const SnapMixin = {
             this._snapList = layers;
         }
 
+        map.off('pm:remove', this._handleSnapLayerRemoval, this);
+        map.on('pm:remove', this._handleSnapLayerRemoval, this);
     },
     _calcClosestLayer(latlng, layers) {
         // the closest polygon to our dragged marker latlng
--- a/src/js/Mixins/Snapping.js
+++ b/src/js/Mixins/Snapping.js
@@ -198,16 +198,16 @@ const SnapMixin = {
 
                 map.off('pm:remove', this._handleSnapLayerRemoval, this);
                 map.on('pm:remove', this._handleSnapLayerRemoval, this);
-
-                // this is for debugging
-                const debugLine = L.polyline([], { color: 'red', pmIgnore: true });
-                debugIndicatorLines.push(debugLine);
-
-                // uncomment 👇 this line to show helper lines for debugging
-                // debugLine.addTo(map);
             }
         });
 
+        // this is for debugging
+        this.debugIndicatorLines = layers.map(() =>
+            L.polyline([], { color: 'red', pmIgnore: true }),
+        );
+
+        // uncomment 👇 this line to show helper lines for debugging
+        // this.debugIndicatorLines.forEach(debugLine => debugLine.addTo(map));
         // ...except myself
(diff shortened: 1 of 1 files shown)


smeijer:
Related PR: #279

I haven't tested, but I don't expect that both are needed. The scope of #279 looks larger by adding MultiPolygon support, whereas this PR exists to "just fix the snapping bug #257".

My first idea of fixing the snapping bug was also to loop over the rings in the `_calcLayerDistances` method. But I decided to take the more simple approach of pre-fixing it in the `_snapList` for performance purpose. (create a couple of extra layers upon init vs an extra loop on every mouse-move).

That being said, maybe it's needed to add support for MultiPoint or MultiLine the same way. But again, haven't tested those. Adding support for them in the same way should be an easy job though.

smeijer:
Turned out, MultiPoint works as expected. Not causing any troubles.

MultiLine lacks the same support. I'll push an update in a few minutes.

smeijer:
Any chance this one can get merged? I'm currently using my own fork. Would be nice if I could use the main branch again.

smeijer:
Rebased on `develop`

codeofsumit:
@smeijer thank you for your work on this.
Just to clarify: if we had proper multi polygon/line support, this would still be needed (as it's part of the solution) - do you agree?

PR #300, opened 2018-06-13
Update TurfJS 6.x - fixes #228

### LINKED ISSUE #228: size is TEN TIMES BIGGER since including turf
turf is the dependency to calculate the necessary things to support the Cutting feature.
Turf thinks about replacing it's big dependency that's causing this size: https://github.com/Turfjs/turf/issues/88

image

Another option would be to replace turf with martinez. Martinez still has some problems though and I wasn't able to make holes work with it due to this: https://github.com/w8r/martinez/issues/44

Afaik, turf thinks about replacing their big dependency `jsts` with martinez too which would be my favorite option.

Let's see 🤞

DenisCarriere:
Update 3 modules from TurfJS to 6.x

Fixes #228

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
   ],
   "main": "dist/leaflet.pm.min.js",
   "dependencies": {
-    "@turf/difference": "^5.1.5",
-    "@turf/intersect": "^6.1.2",
-    "@turf/kinks": "^5.1.5"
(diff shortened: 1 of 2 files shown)


codeofsumit:
amazing! Thank you @DenisCarriere and the whole turf team!
image

PR #306, opened 2018-06-21
Increase GeoJSON precision

gorunovanton:
I've increased precision when working with GeoJSON due to false-positive polygon self-intersection in scenarios with two close points.

Dataset with false-positive self-intersection:
[false_self_intersection.txt](https://github.com/codeofsumit/leaflet.pm/files/2123561/false_self_intersection.txt)

### DIFF
--- a/src/js/Draw/L.PM.Draw.Cut.js
+++ b/src/js/Draw/L.PM.Draw.Cut.js
@@ -24,7 +24,7 @@ Draw.Cut = Draw.Poly.extend({
             // only layers with intersections
             .filter((l) => {
                 try {
-                    return !!intersect(layer.toGeoJSON(), l.toGeoJSON());
+                    return !!intersect(layer.toGeoJSON(15), l.toGeoJSON(15));
                 } catch (e) {
                     console.error('You cant cut polygons with self-intersections');
                     return false;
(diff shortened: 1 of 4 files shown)


codeofsumit:
oh wow, thanks for your work on this! I can't seem to find documentation about the parameter passed to `toGeoJSON` - can you provide a link?

themre:
very nice! perhaps put this in some constant so it can be override via settings?

codeofsumit:
great idea @themre

themre:
found reference to this precision parameter. we need to make sure from which version is this available:
https://github.com/Leaflet/Leaflet/blob/88dd22cfe89a209d19c061a813e99669844a921b/src/layer/GeoJSON.js#L292

gorunovanton:
There is no information about a precision argument in the official Leaflet documentation of the .toGeoJson() method.

But i've found a pull request with it which was merged around one year ago into the Leaflet codebase:
https://github.com/Leaflet/Leaflet/pull/5444

I don't think that putting this value into a constant is necessary due to how floating point operations are implemented in JS - JS always operates on FP numbers as double precision numbers and 15 is the amount of most significant digits in double precision numbers. This opinion based on two SO topics:
https://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript/3644302#3644302
https://stackoverflow.com/questions/28045787/how-many-decimal-places-does-the-primitive-float-and-double-support

gorunovanton:
found that support of this parameter was added in Leaflet 1.1.0:
https://github.com/Leaflet/Leaflet/tree/v1.1.0

PR #313, opened 2018-07-09
Circle Options are now properly handled after draw (Patch)

codeofsumit:


codeofsumit:
note to myself: add this as a test:
https://jsfiddle.net/n3gL6wty/

PR #314, opened 2018-07-10
Snap middle of segments

KenAlin:
Hi,

This PR add an option, disabled by default, allowing snap middle of segments.

Here is a quick demo : 

leafletpm snap middle

Usage : enable with `snapMiddle: true` in your options. (See edited README)

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ var options = {
     snappable: true,
     snapDistance: 20,
 
-    // allow snapping with the middle of segments
+    // allow snapping to the middle of segments
     snapMiddle: false,
 
     // self intersection
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ map.pm.enableDraw('Poly', { finishOn: 'dblclick' });
 map.pm.disableDraw('Poly');
 ```
 
-All available options are specified in the Drawing Mode Section below
+All available options are specified in the Drawing Mode Section below.
 
 ##### Drawing Mode
 
@@ -129,6 +129,9 @@ var options = {
     // snapping
     snappable: true,
     snapDistance: 20,
+
+    // allow snapping with the middle of segments
+    snapMiddle: false,
 
     // self intersection
     allowSelfIntersection: true,
--- a/src/js/Mixins/Snapping.js
+++ b/src/js/Mixins/Snapping.js
@@ -1,7 +1,8 @@
+import Utils from '../L.PM.Utils';
+
 const SnapMixin = {
     _initSnappableMarkers() {
         this.options.snapDistance = this.options.snapDistance || 30;
-        this.options.snapMiddle = this.options.snapMiddle;
 
         this._assignEvents(this._markers);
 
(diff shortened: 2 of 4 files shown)


codeofsumit:
cool! Looks very good already from feature and code 👍 . I'll be on vacation tomorrow for a week so I have to postpone this PR until then but huge thanks for this contribution. Looking forward to merge it 👊 <3

README.md
```diff
@@ -130,6 +130,9 @@ var options = {
     snappable: true,
     snapDistance: 20,
 
+    // allow snapping with the middle of segments
```
codeofsumit: I'd rather say `...snapping to the middle...`

src/js/Mixins/Snapping.js
```diff
@@ -1,6 +1,7 @@
 const SnapMixin = {
     _initSnappableMarkers() {
         this.options.snapDistance = this.options.snapDistance || 30;
+        this.options.snapMiddle = this.options.snapMiddle;
```
codeofsumit: is this necessary? The line above was only to create a default of 30 🤔

src/js/Mixins/Snapping.js
```diff
@@ -149,10 +150,19 @@ const SnapMixin = {
...
+        let shortestDistance = distanceAC < distanceBC ? distanceAC : distanceBC;
+
+        // snap to middle of segment if enabled
+        if (this.options.snapMiddle) {
+            const M = this._middleLatLng(map, A, B);
+            const distanceMC = this._getDistance(map, M, C);
+            const middleIsNearest = (distanceMC < distanceAC && distanceMC < distanceBC);
+            closestVertexLatLng = middleIsNearest ? M : closestVertexLatLng;
+            shortestDistance = middleIsNearest ? distanceMC : shortestDistance;
+        }
```
codeofsumit: should we add a few more explanation commends between those lines to explain what exactly is done inside the if-clause?

src/js/Mixins/Snapping.js
```diff
@@ -333,6 +343,12 @@ const SnapMixin = {
     _getDistance(map, latlngA, latlngB) {
         return map.latLngToLayerPoint(latlngA).distanceTo(map.latLngToLayerPoint(latlngB));
     },
+    _middleLatLng(map, latlngA, latlngB) {
+        const p1 = map.project(latlngA);
+        const p2 = map.project(latlngB);
+
+        return map.unproject(p1._add(p2)._divideBy(2));
```
codeofsumit: ha, nice one!. Did not know `_divideBy(2)`

codeofsumit:
I added a few comments and will think about in what way this could be testable via cypress 🤔 . I haven't added any snapping tests yet so I might start with them too 👍

src/js/Mixins/Snapping.js
codeofsumit: are these private functions that we rather shouldn't use or would you say it's ok?

KenAlin:
Thanks for your review !
I will commit some changes according to your comments.

src/js/Mixins/Snapping.js
KenAlin: In fact, I reused the method `_calcMiddleLatLng` (see [here](https://github.com/codeofsumit/leaflet.pm/blob/develop/src/js/Edit/L.PM.Edit.Line.js#L565)) in `L.PM.Edit.Line.js`. But with next commit, it will be moved into an "Utils.js" file.

About `_add()` and `_divideBy()` : these are used by public functions `add()` and `divideBy()` (see [Leaflet sources](https://github.com/Leaflet/Leaflet/blob/9fb0df7f0418243dd50ffaac715dc894a7afeb91/src/geometry/Point.js#L48)), but are destructive. That's fine here, because p1 and p2 are used only to calculate middle of segment. 😊

src/js/Mixins/Snapping.js
KenAlin: Removed

codeofsumit:
Alright @KenAlin this looks very good. Nice code, nice code comments, it's all there 👍
Thanks a ton for your contribution! Happily merged!

PR #321, opened 2018-07-19
Markercluster Compatibility

### LINKED ISSUE #316: [Bug] Problem with PM and MarkerCluster
Hi Sumit :)

First of all, thank you for your great contribution of this awesome plugin for Leaflet. I really appreciate the work you do here.

I have the following problem with the combination of Leaflet.PM and Leaflet.markercluster:

- The edit mode is disabled for a single marker by `pmIgnore: true` -> so far so good
- If I remove one marker, all of them are getting deleted -> that should not happen, especially because I don't wont to be able to delete marker at all.

Fiddle to reproduce:   http://jsfiddle.net/tc7xwqz0/20/

In the end, I'm not sure if it is a problem of PM or MarkerCluster. Perhaps you can have a look and let me know what you think.

Best regards,
Jan

codeofsumit:
This PR is aiming to make leaflet.markercluster work with leaflet.pm. Fixes #316.

Current status:
It's hard to detect layers from markercluster. Will create an issue in their repo and ask for an identification property.

ToDos:

- [ ] toggleGlobalEdit Mode option passing has to work for markers that are added by markercluster later
- [ ] removal mode has to work, markercluster should update correctly after removals

### DIFF
--- a/cypress/integration/marker.spec.js
+++ b/cypress/integration/marker.spec.js
@@ -27,29 +27,34 @@ describe('Draw Marker', () => {
             const markerCluster = L.markerClusterGroup([]);
             console.log(markerCluster.options);
 
-            markerCluster.addLayer(L.marker([0, 1]));
+            // map.pm.toggleGlobalEditMode({
+            //     draggable: false,
+            // });
+
+            const marker1 = L.marker([0, 1]);
+            markerCluster.addLayer(marker1);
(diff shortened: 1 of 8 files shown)


codeofsumit:
I've opened the issue at the markercluster repo
https://github.com/Leaflet/Leaflet.markercluster/issues/905

PR #324, opened 2018-07-27
[bugfix] Fix snapping for single point lines/poly

KenAlin:
Hi,

This PR aims to fix a bug detected in snapping. After drawing a "single point line" (a line where begin = end), console errors occurs.

leafletpm bug snapping


codeofsumit:
Hi @KenAlin, thanks so much for your contribution. I'm thinking of fixing this in the way that a PolyLine is removed when it has only one coordinate because that's how we do it during vertex deletion as well.
Also you can't add another vertex to it as it's just one marker when editing.

What do you think?

codeofsumit:
closed in favor of #349

PR #338, opened 2018-09-02
Geojson cut

codeofsumit:
GeoJSON Cutting has a problem. This branch aims to fix it.
Might be caused by turf/difference, not 100% sure though.
Turf Issue: https://github.com/Turfjs/turf/issues/1475

The Problem:
cutproblem


codeofsumit:
I have tried to downgrade to @Turf 5.x to no avail. Apparently, in that version the problematic shape is not supported/ignored completely.

codeofsumit:
Most promising fix for this is [Turf v7](https://github.com/Turfjs/turf/pull/1432). I would wait for it.

codeofsumit:
Closed in favor of https://github.com/codeofsumit/leaflet.pm/pull/403

PR #367, opened 2018-12-12
Fix selfIntersection check on cut

bugleev:
Found a bug, steps to reproduce:
1. Draw a polygon
2. Start drawing a cutting shape in it with allowSelfIntersection = false

selfintersection-false

3. Click on first vortex (as the cursor snaps to it)
4. Shape is cut

selfintersection-done


codeofsumit:
Thank you for your work on this, @bugleev 👍

PR #390, opened 2019-01-18
Drag Mode: Create Separate Drag Button

### LINKED ISSUE #379: [Feature Suggestion] Use a dedicated button to move layers
As reported also in https://github.com/r-spatial/mapedit/issues/86#issuecomment-452325984, 
 when in editing mode on polygons clicking nearby but not really "over" a vertex and then moving the mouse (or  trying to pan on the map by clicking and dragging to move to another area) can lead to moving  the whole polygon "by mistake". 
I think therefore that it would be better if the "move whole polygon" functionality could be implemented using a "dedicated" button. As an alternative, it could be implemented by allowing to move the whole polygon only by clicking on its centroid, as implemented [here](http://leaflet.github.io/Leaflet.draw/docs/examples/full.html) (for leaflet.draw) for rectangles and circles.

codeofsumit:
- [x] Drag Lines
- [x] Drag Polygons
- [x] Drag Markers
- [x] Drag Circles
- [x] Drag Rectangles
- [x] Markers and Circles can still be dragged in edit mode due to their nature
- [x] Add New Drag Icon
- [x] Add New Edit Icon as Edit Mode becomes strictly about editing Vertices
- [x] Check if the `preventVertexEdit` option is still needed

image

Will fix #379.


TravnikovRN:
It's so rude decision to separate

PR #400, opened 2019-02-11
add option to ignore snapping to some of layers

AliEmreYilmaz61:
add "snapIgnore:true" in a layers option to ignore it during snapping


src/js/Mixins/Snapping.js
```diff
@@ -212,15 +212,19 @@ const SnapMixin = {
...
+          }
+
 
         // this is for debugging
         const debugLine = L.polyline([], { color: 'red', pmIgnore: true });
         debugIndicatorLines.push(debugLine);
 
         // uncomment 👇 this line to show helper lines for debugging
-        // debugLine.addTo(map);
+         //debugLine.addTo(map);
```
codeofsumit: Super small thing, but please revert this change ;-)

AliEmreYilmaz61: done :)

AliEmreYilmaz61:
Updated Readme. Looked a bit for testing, not possible to trigger mouse events with Cypress, would be happy to hear if you got any ideas.

src/js/Mixins/Snapping.js
```diff
@@ -212,8 +212,11 @@ const SnapMixin = {
         layer instanceof L.Polyline ||
         layer instanceof L.Marker ||
         layer instanceof L.CircleMarker
+
       ) {
-        layers.push(layer);
+          if(layer.options.snapIgnore !== true){
```
codeofsumit: Could this condition be part of the above if-clause in line 211?

README.md
```diff
@@ -217,6 +217,16 @@ Here's a list of layer events you can listen to:
 | pm:unsnap       | `e`    | Fired when a vertex is unsnapped. Payload is the same as in `snapdrag`                                               |
 | pm:centerplaced | `e`    | Called when the center of a circle is placed/moved.                                                                  |
 
+For making the snapping to other layers selective, you can add the "snapIgnore" option to your layers to disable the snapping to them during drawing.
+```js
+L.geoJSON(data,{
+  style : this.style,
```
codeofsumit: `this.style` seems like specific to your application. I would remove that but what do you think?

codeofsumit:
You're right regarding the cypress tests - I don't have any test for snapping yet because of the same reason. If you can fix the last two comments I have, update the branch, I'll merge it ✅

src/js/Mixins/Snapping.js
AliEmreYilmaz61: Fixed

README.md
AliEmreYilmaz61: Correct, I removed it

codeofsumit:
Great job 👏

PR #403, opened 2019-02-15
Fix GeoJSON Cut

codeofsumit:
This replaces the PR #338 which was too old and has too many merge conflicts.

Turf Issue: [Turfjs/turf#1475](https://github.com/Turfjs/turf/issues/1475)

The Problem:


codeofsumit:
I tried polygon-clipping https://github.com/mfogel/polygon-clipping but it removed holes when cutted. It also adds some transformational overhead as it doesn't accept geojson but only the `coordinates` array.

Falke-Design:
@codeofsumit this does not happen any more if all steps are done from a clean map. 
With your Geojson it doesn't work but I think it is not correct / representative. 

Old behavior: https://jsfiddle.net/falkedesign/soLwpt6n/
New:  https://jsfiddle.net/o1dwu2vg/

1.  Draw a Polygon
2. Cut it in two shapes
3. Cut a hole
4. Cut another hole

PR #406, opened 2019-02-19
Release 3.0

codeofsumit:
This release branch aims to collect breaking changes until we can release the next major version.
It has to constantly stay in sync with the `develop` and `master` branch to avoid extensive merge conflicts down the road.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ See the available options in the table below.
 | snapDistance          | `20`                                  | the distance to another vertex when a snap should happen                                                                                              |
 | snapMiddle            | `false`                               | allow snapping to the middle of a layers segments (between two vertexes)                                                                              |
 | tooltips              | `true`                                | show helpful tooltips for your user                                                                                                                   |
-| allowSelfIntersection | `true`                                | allow self intersections                                                                                                                              |
+| allowSelfIntersection | `false`                               | allow self intersections                                                                                                                              |
 | templineStyle         | `{ color: 'red' },`                   | [leaflet path options](https://leafletjs.com/reference-1.4.0.html#path) for the lines between drawn vertices/markers.                                 |
 | hintlineStyle         | `{ color: 'red', dashArray: [5, 5] }` | [leaflet path options](https://leafletjs.com/reference-1.4.0.html#path) for the helper line between last drawn vertex and the cursor.                 |
 | cursorMarker          | `true`                                | show a marker at the cursor                                                                                                                           |
@@ -224,7 +224,7 @@ Let's you edit vertices of layers. Use it like this:
 ```js
 // enable edit mode
 layer.pm.enable({
-  allowSelfIntersection: false,
+  allowSelfIntersection: true,
 });
 ```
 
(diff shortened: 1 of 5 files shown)


Falke-Design:
Should be also removed:

L.PM.Draw.Marker, L.PM.Draw.CircleMarker:
```
this._map.fire('pm:create', {
      shape: this._shape,
      marker, // DEPRECATED
      layer: marker,
    });
```

L.PM.Draw.Line:
```
    // fallback option for finishOnDoubleClick
    // TODO: remove in a later release
    if (this.options.finishOnDoubleClick && !this.options.finishOn) {
      this.options.finishOn = 'dblclick';
    }
```

L.PM.Map:
```
enableDraw(shape = 'Polygon', options) {
    // backwards compatible, remove after 3.0
    if (shape === 'Poly') {
      shape = 'Polygon';
    }
   //...


disableDraw(shape = 'Polygon') {
    // backwards compatible, remove after 3.0
    if (shape === 'Poly') {
      shape = 'Polygon';
    }
   //...
```

Falke-Design:
Also old button nams in L.PM.Toolbar in `addControls()` and `toggleButton()`

```
    // make button renaming backwards compatible
    if (typeof options.editPolygon !== 'undefined') {
      options.editMode = options.editPolygon;
    }
    if (typeof options.deleteLayer !== 'undefined') {
      options.removalMode = options.deleteLayer;
    }
```

Falke-Design:
Thinking of changing the event `markerdragstart` into `vertexdragstart` same for the other "marker" events. It would be a clearer name

Falke-Design:
Remove the first argument `e`, because it is also in the object:
L.Controls:
```
  _triggerClick(e) {
    // TODO is this a big change when we change from e to a object with the event and the button? Now it's the second argument
    this._button.onClick(e, { button: this, event: e });
    this._clicked(e);
    this._button.afterClick(e, { button: this, event: e });
  },
```

Falke-Design:
Thinking of rename Line to Polyline. To match the leaflet default

Falke-Design:
Change the option name

editable | false | makes a CircleMarker editable like a Circle
-- | -- | --

Falke-Design:
Remove the optional `shape` it has no function, all shapes will be disabled: 
https://github.com/geoman-io/leaflet-geoman/blob/ff58d546ae5807a331485c16b5ac442b6bbce527/src/js/L.PM.Map.js#L62-L69

Falke-Design:
Mode.Edit, Mode.Drag, ... add `_` to private functions

Falke-Design:
Make `_removeLastVertex()` to a public function

Falke-Design:
L.PM.Map and all Modes change `this.map` to `this._map`. Then it is used the same name as in the Draw and Edit Classes

Falke-Design:
rename `_finishShape()` and  `_createMarker` to the same name and make it public

Falke-Design:
Add to all CSS classes the prefix `leaflet-geoman`. More and more libraries have problems with using the same CSS-Class name. Like `.active`

Falke-Design:
Draw.Line uses `templineStyle`, Draw.Rectangle uses `pathOptions`.

PR #436, opened 2019-05-11
Added shift holding option 2.0

Falke-Design:
Add 45° steps for Lines, Polys and Cutting on shift holding. Rectangle is a square if shift is holding. Option allowShift added.

I had problems with the listener and "this". So I used window vars. Maybe you have a better idea.

Created a branch and updated the code with own function and removed, from me added, turfs functions.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -333,7 +333,7 @@ polygonLayer.on('pm:intersect', function(e) {
 // });
 // map2.pm.disableGlobalEditMode();
 
-map2.pm.enableDraw('Polygon', { allowSelfIntersection: false });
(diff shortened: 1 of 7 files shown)


ianjanicki:
Duplicated these changes in my fork; can confirm this works great. Thanks @Falke-Design.

I use [L.CRS.Simple](https://leafletjs.com/reference-1.5.0.html#crs-l-crs-simple) (which is a cartesian projection) as my coordinate system and that throws everything off unfortunately. Currently working on a fix for that.

Falke-Design:
Created subplugin [PMOrtho](https://github.com/Falke-Design/PMOrtho)

PR #447, opened 2019-05-25
Triangle

Falke-Design:
Add triangle draw function


Falke-Design:
Created subplugin [GeomanTriangle](https://github.com/Falke-Design/GeomanTriangle)

Issue #448, opened 2019-05-28
Draw Rectangle snap target misplaced


ianjanicki:
Repro:
* [Use demo site](https://leafletpm.now.sh)
* Snap new Polygon vertex to existing point
   * Vertex snaps directly on existing point
* Snap closing Rectangle vertex to existing point
   * Vertex **does not** snap to existing point, but to current lat,lng of mouse

Is this desired behavior? Because it seems inconsistent w/ Polygon behavior.

**[Giphy of repro steps](http://www.giphy.com/gifs/hVlR2cY3dwd4vr6J0a)**

**Edit:** I just noticed that this behavior isn't present in Edit mode, just the closing click of Draw Rectangle

ianjanicki:
After testing locally, found the issue in `Draw.Rectangle`, `_finishShape(e)` method:

`const B = e.latlng;`

Should be:
`const B = this._hintMarker.getLatLng();`

Let me know if best practice for this project is to push the fixed branch.

PR #449, opened 2019-05-30
Fix Rectange snap on Draw

ianjanicki:
When drawing a Rectangle, and the user wants to snap to a nearby vertex, this fix makes sure it lands on the vertex vs. the current x,y of the cursor.


codeofsumit:
Thanks for the PR 🎉 . Does this still work on mobile for snapping? Can you verify this?

codeofsumit:
@ianjanicki thanks for checking. Could you also test it on a real mobile device? Because on mobile there are other events used inside leaflet and it affects the behaviour of this lib

ianjanicki:
@codeofsumit Would love to! But have a few questions about doing that:
1. Your `CONTRIBUTING.md` is a little confusing/misleading. `/index.html` does not work for me; it's a blank page. Are you referencing `/demo/index.html`? This works on my machine.
2. I was trying to use Node's `http-server` to host `demo/index.html` so I could test it on my phone (using an Android phone + Mac) — but the `addControls` method fails, so there's no Leaflet.pm toolbar, so that was a dead-end.
3. I noticed you host on now.sh, and I'm curious if you host for every branch — if this PR has a build URL would gladly test it there.

Let me know if I'm doing anything wrong here.

codeofsumit:
Sorry @ianjanicki, accidentally fixes this already through #448.
But I took your feedback and made changed to the contributing.md and hope it makes things clearer where and how to manually test your changes on a working map 👍
Please open another issue in case contribution is still hard/confusing. Thank you for your time and work!

PR #460, opened 2019-06-27
pmLock option

Falke-Design:
With pmLock layers can be locked and unlocked for editing.

Simply change the layer options to lock the layer: 
`layer.options.pmLock = true;` 
and unlock:
`layer.options.pmLock = false;` 

or: 
`map.pm.enableDraw('Rectangle', { pathOptions:{ pmLock: true }});`
`map.pm.enableDraw('Marker', { markerStyle:{ pmLock: true }});`


Falke-Design:
Created subplugin [PMLock](https://github.com/Falke-Design/PMLock)

PR #464, opened 2019-07-06
emit an event while the corner of a rectangle is being dragged

jgravois:
greetings! 👋 

I'm updating a [**very** old](http://projectionwizard.org) leaflet application that invokes some special logic while the corner of a rectangle is being dragged to prohibit its span from surpassing 360°.

emitting the event below did the trick so I figured I'd propose it as a patch.


jgravois:
>  it should be added to all shapes and referenced inside the README

done (and done).

jgravois:
if there's anything on my side that's keeping this from being landed, just say the word.

codeofsumit:
closed in favor of #489

PR #468, opened 2019-07-12
Create Indonesian language translation

andyprasetya:


codeofsumit:
Hi and thanks for the pull request. To make the language work, you also have to import the file here: https://github.com/codeofsumit/leaflet.pm/blob/develop/src/assets/translations/index.js

codeofsumit:
Oh and please add it do the documentation as well

andyprasetya:
Sure, I'll do it.

PR #472, opened 2019-07-17
Drag Layergroup

Falke-Design:
Add drag function for layergroups.
To drag a layergroup, it must have the option **pmDrag = true**.
But only L.FeatureGroup or L.GeoJSON are working, because of this: [https://github.com/Leaflet/Leaflet/issues/4861](https://github.com/Leaflet/Leaflet/issues/4861)

New map functions:
`toggleLayerGroupDragMenu(); //To change the menu items from drag (Layer,Layergroup,Cancle) || (Cancle)
changeDragMode(mode); //Change between dragging layers (0) or layergroups (1)`

Also added a fallback for not existing translations.

In L.Controls I change the onRemove function, because if it was called the buttons are vanished but the containers and modes are active / true.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -1,30 +1,18 @@
 /* eslint-disable */
 // Provide your access token
-const accessToken =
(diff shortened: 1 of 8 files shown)


Falke-Design:
Created subplugin [PMDrag](https://github.com/Falke-Design/PMDrag)

PR #473, opened 2019-07-23
Add Indonesian language translation to index.

andyprasetya:


src/assets/translations/index.js
```diff
@@ -1,5 +1,6 @@
 import en from './en.json';
 import de from './de.json';
+import de from './id.json';
```
codeofsumit: variable name should be `id`, not `de`

codeofsumit:
Oh btw, please do this in your original PR by commiting to that branch
https://github.com/codeofsumit/leaflet.pm/pull/468

PR #489, opened 2019-09-09
Added MarkerDrag & extended MarkerDragStart

bumbummen99:
Added markerdrag event to Line editor.
Added workingLayer to markerdragstart in Line editor.
Increased version
Improved some comments


src/js/L.PM.Utils.js
```diff
@@ -9,4 +9,4 @@ const Utils = {
   },
 };
 
-export default Utils;
```
codeofsumit: please remove this change

src/js/L.PM.js
```diff
@@ -108,4 +108,4 @@ L.PM = L.PM || {
 };
 
 // initialize leaflet.pm
-L.PM.initialize();
```
codeofsumit: please remove this change

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "leaflet.pm",
-  "version": "2.2.0",
```
codeofsumit: Please do not update the version. It will be bumped when I release a new version of the library.

src/js/L.PM.Utils.js
bumbummen99: The change has been removed

src/js/L.PM.js
bumbummen99: The change has been removed

package.json
bumbummen99: Done

bumbummen99:
Sorry for the long silence, the requested changes have been made. I am not sure but i think it would be convenient to add the current latlng and marker to the event. Also this might need tests.

Falke-Design:
close #663 is already merged

PR #495, opened 2019-09-23
Tiny PR: Use container class instead of button config to toggle control

limfinity:
This seemed like a bug to me, as the toggling did not work on re-renders of my application.


codeofsumit:
@limfinity thanks for the PR. Can you explain a bit more which problem it solves? Or ideally, show a JSfiddle that shows the problem?

PR #504, opened 2019-10-04
Added opt-in Functionality

### LINKED ISSUE #191: Is it possible to set pmIgnore=true by default?
Its a bit too inconvenient to set this option for each layer created, because sometime it could be created in 3rd party plugin. It would be great if I can set pmIgnore = true for all layers, and then select some layers to be handled by PM.

### LINKED ISSUE #439: Opt-in functionality
I'm finding difficult to restrict the PM edit/drag/remove functionality to specific layers in my map. In other words, difficulty to exclude other layers (in particular, those created by third-party tools) where the PM functionality is not required at all. The `pmIgnore` mechanism is not really useful in these cases as it would minimally require one to traverse those layers and do something only for purposes of disabling PM functionality for them.

The kind feature request here is to provide an initialization option to indicate the desire for the PM functionality in an *opt-in* fashion. More concretely, something like the following:

- An new option `optIn` (or some other name for the intended effect) for the `addControls` method, of course with default value `false` for backwards compatibility.

- When adding a particular layer in my application, be able to indicate `pmEnable` (with basically the opposite meaning of `pmIgnore`). Example:

    ```js
    L.marker([51.50915, -0.096112], { pmEnable: true }).addTo(map);
    ```

Imho, this would really make your wonderful tool even more appealing!  Thanks!

codeofsumit:
This PR adds a possibility to initiate leaflet-geoman with an optIn option. Closes #439, Closes #191 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,21 @@ their options when creating them. Example:
 L.marker([51.50915, -0.096112], { pmIgnore: true }).addTo(map);
 ```
 
+##### Opt-In
+
+If you want to use leaflet-geoman as opt-in, call the following function right after importing:
+
+```js
+L.PM.initialize({ optIn: false });
+```
+
+All layers will be ignored by leaflet-geoman, unless you specify `pmIgnore: false` on a layer:
+
+```js
+L.marker([51.50915, -0.096112], { pmIgnore: false }).addTo(map);
+```
+
+
 #### leaflet-geoman Toolbar
 
 <img align="left" height="200" src="https://file-ffrjxxowri.now.sh/" alt="leaflet-geoman Toolbar">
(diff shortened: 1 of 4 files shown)


limfinity:
Hey @codeofsumit many thanks for tackling this issue so fast! 💯 

I tried it out and looked at the changes and would have some suggestions. While trying it out I saw, that when you do not pass any `optIn` value, the `pm` object is missing on existing layers.

Also when I add the `optIn` value I would have expected that layers created with geoman itself would have the `pmIgnore` flag set by default to `false`. Is there a way I could do that myself? Because when I create a marker in this way for example, it can't even create them because it is trying to call `enable` on the undefined `pm` object.

In general the whole wording and combination of `optIn` and `pmIgnore` was hard to understand and makes the code much more verbose than I thought.

codeofsumit:
@limfinity you're right. Do you mind opening up a new issue explaining this again (or just copy paste)? I will tackle it asap

PR #511, opened 2019-10-16
Replace `optIn` with a `useType`

limfinity:
Hi @codeofsumit! 
In order to make your wonderful library work for me and to show our needs better I create this **Draft** PR. I started working on this, because my performance was significantly reduced. And I believe the root cause is the init hooks, that get added to every Leaflet layer. (I have a map with more than 10k Leaflet elements.)

As my changes change the mechanics of the library quite a bit, I want to include you as early as possible. And would love to hear if this makes sense to you and how we should work from here.

Short description of the changes. Before we had two attributes on different layers that were able to toggle the abilities of the library. That was `optIn` on a more global level and `pmIgnore` on a more local (layer) level. I found this concept hard to grasp and also as [commented on your initial PR](https://github.com/geoman-io/leaflet-geoman/pull/504#issuecomment-540600338)  #504 had some problems to make this work.
So to make the handling easier I introduced a `useType` which currently can have two values
* `ALL_LEAFLET_LAYERS` use geoman on all available Leaflet layers/objects 
* `GEOMAN_LAYERS_ONLY` use geoman only on layers drawn with geoman itself so only available to extended Leaflet classes

This obviously removes the possibility to enable geoman on a specific layer, as you could have done before. Adding this functionality could be done with another `useType` and handling it accordingly.

Another note: I haven't checked all use cases by far, and so it is very well possible that parts are not working as expected. I just want to get your thoughts first. Before investing more time.

Appreciate your time and looking for your feedback! 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@
 - [Customization](#customize)
 - [Need a feature?](#feature-request) | [Existing Feature Requests](https://github.com/geoman-io/leaflet-geoman/issues?q=is%3Aissue+is%3Aclosed+label%3A%22feature+request%22+sort%3Areactions-%2B1-desc)
 
-
 ### Installation
 
 #### Migrate from Leaflet.PM
@@ -51,6 +50,7 @@
 npm uninstall leaflet.pm
 npm i @geoman-io/leaflet-geoman-free
 ```
+
 That's it.
 
 #### Install via npm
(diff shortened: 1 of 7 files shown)


codeofsumit:
hi @limfinity, hm that option currently does not seem like a direction I would like to go.
I will think about it a bit more. Nonetheless, the `optIn` part is correct and I'm going to change that and performance is next on my list to tackle anyway. If you have large geojson files to test against, send them my way :-)

codeofsumit:
I'm going to close the now as I have made some progress on both fronts. The OptIn situation will be completely revamped and new options will be introduced to handle massive amounts of that. Will post updates in all related Issues/PRs once I can show something

PR #524, opened 2019-12-05
Fixes error when providing a "weight" value to hintLine style (leaflet 1.4.0+)

cosme-benito:
- Added workaround for #443

- Updated examples to leaflet 1.6
- Added "weight" to a hintLine in a cypress test


cosme-benito:
The problem is that whenever you set a weight value to an empty Line/Polygon (ie: with no vertex added yet) Leaflet just throws an exception, which leaves geoman in an inconsistent state.  

I simply added an if so it doesn't apply the style to an empty line/polygon. As far as I can tell there are no side-effects. When you add a new vertex the hintstyle will be re-applied anyway.

I just changed from

```javascript
if (this._doesSelfIntersect) {
      this._hintline.setStyle({
        color: 'red',
      });
    } else {
      this._hintline.setStyle(this.options.hintlineStyle);
    }
```

to

```javascript
if (this._doesSelfIntersect) {
      this._hintline.setStyle({
        color: 'red',
      });
    } else if (!this._hintline.isEmpty()) {
      this._hintline.setStyle(this.options.hintlineStyle);
    }
```

Issue #493, opened 2019-09-20
Edit Mode - Keep Snapped Items Together


solutegrate:
I have read through everything and I can't find a way to keep items together when moving an marker.

Example:

I have (2) Lines snapped to (1) Marker. I would love a keyboard shortcut to:

1. I hold down ex. "Ctrl"
2. Click and drag the marker
3. The (2) lines snapped to the marker should move with the marker.

I hope this make sense.

codeofsumit:
Yes this makes perfect sense. I'm thinking about this already 👍 . I'll flag your issue as a Feature Request and close the issue. Others can upvote this which helps me prioritize it but it's definitely something I want to build soon - thanks for bringing it up!

solutegrate:
Awesome! By the way, this plugin is bad ass! I'll be contributing financially for sure. Thanks for everything and can't wait for this feature to be available.

Although, I did think about this:

When dragging the marker with the lines snapped to it, would it drag from the previous vertex of the line or would it drag from the first vertex of the line?  This one option could render different results.

codeofsumit:
Take a look at this:
![pinning](https://user-images.githubusercontent.com/2399810/65375984-288ece80-dc9b-11e9-930e-bca03ad7cb56.gif)

bumbummen99:
@codeofsumit This looks kind of the same as my WIP right now, will that be the pinning feature of the pro version?

cjauvin:
@bumbummen99 I'd be really interested to know more about your WIP, is it #489 ?

bumbummen99:
@cjauvin It is also based on that PR but the logic behind it is a bit more complex and not included. Basically, i store all layers for snapped dragging into a dedicated LayerGroup. When i drag a layer, i simply iterate all layers in that group and check their latlang or latlangs for matching/snapped ones. I then check if they are the same with some precision offset. If i find any matches i disable snapping for them (important) and update their respective latlang on marker drag. You will have to disable snapping during dragging as you would just snap your dragging layer to the layers you would want to drag too constantly. Also editing has to be disabled as it will spawn additional markers (the handles). On drag end i simply take the temp list of snapped layers and re-enable snapping and editing for them. If your layers are not editable all the time you can skip disabling it.

I think the solution itself is not perfect nor pretty and it will be slower with an increasing amount of layers, I also still have to figure out a way to circumvent this limitation and maybe get rid of the iteration needed to find snapped layers.

Maybe @codeofsumit also has an idea on it

[...]

PR #537, opened 2020-02-03
Add edit option to keep snapped vertices together

cjauvin:
Following this [discussion](https://github.com/geoman-io/leaflet-geoman/issues/493) I have implemented an option for the edit tool to keep the snapped vertices together when dragging:

geoman_pr2

Note that this currently works for polygons and lines, but not for other shapes. I have also experimented with a mechanism allowing the "shared" middle markers to be moved together, but since it's more complex and has additional `turf` dependencies, I'm not including it in this PR.

If there's any interest in the feature I will add some related tests.

### DIFF
--- a/cypress/integration/line.spec.js
+++ b/cypress/integration/line.spec.js
@@ -112,11 +112,11 @@ describe('Draw & Edit Line', () => {
 
     // draw a line
     cy.get(mapSelector)
-      .click(90, 250)
-      .click(100, 50)
-      .click(150, 50)
-      .click(150, 150)
-      .click(150, 150);
+      .click(390, 250)
+      .click(400, 50)
+      .click(450, 50)
+      .click(450, 150)
+      .click(450, 150);
 
     // button should be disabled after successful draw
     cy.toolbarButton('polyline')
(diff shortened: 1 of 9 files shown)


codeofsumit:
Hello all,

@cjauvin I really appreciate your work on this! A healthy community contribution is vital to Leaflet and this library. I know this took you some time, so I'm even more sorry to say that I have different plans for this feature in both, the implementation and the distribution. I have worked on this feature myself for the last weeks.

Over the past months where I haven't been active here, I have been talking to businesses that use Leaflet-Geoman extensively in their teams and projects. As I have reiterated in another issue, I want to dedicate more time to this library and a `Pro` version is exactly how I want to do this.
*More advanced features in a paid version that will trickle down to the open source library over time.*.

Pinning Shared Vertexes is the first feature I will offer in the Pro version of Leaflet-Geoman and I'm actively working on it right now.
In the GIF below, you can see Pinning and Snapping available via an "Options Toolbar" (also exclusive to the Pro version for now) and Pinning works exactly as described here.

options-toolbar

If the Pinning options is set on a layer, any vertex you drag from that layer will move any other vertex from any other layer that is equal to it.

If you are interested in Leaflet-Geoman Pro ⭐, consider becoming a [Github Sponsor and Leaflet-Geoman Supporter](https://github.com/sponsors/codeofsumit).

Eventually, this feature will be merged from the Pro version to the OSS version. Until then, this PR will always be here (albeit closed) and as long as @cjauvin is not removing his fork, it will be available to download.

I will post more updated about this in this issue: https://github.com/geoman-io/leaflet-geoman/issues/493

PR #548, opened 2020-02-23
Handle intersection on first marker at polygon

Falke-Design:
Now the polygon can't be finished if clicking on the first marker when it intersects. 
The problem is, that the _finishShape is called from the first placed marker onclick event. So it never called _createVertex and it was never checked if the new latlng was intersected.
If the parameter is true then the new vertex will added to the layer and checked if it intersects.


codeofsumit:
What happened here? 😅

Falke-Design:
I accidentally deleted the Branche.
I fixed the issue, that a polygon can be finished when it intersects:


Added in a new commit a test

Issue #552, opened 2020-03-04
Bug: Circle with CRS.Simple, radius infinite


beig:
When using a CRS.Simple map, the circle doesn't work.
A circle is added with an infinite radius I think. Don't know how it can be properly described 

https://jsfiddle.net/komzs2fx/

beig:
aplogies for bothering you @codeofsumit 
could I get an estimate on my PR?

codeofsumit:
will check it this week 👍

wggley:
Any status?

Falke-Design:
@wggley @beig we merged this into the develop, it will be in the next release

PR #553, opened 2020-03-05
use map's crs to calculate distance

beig:
https://github.com/geoman-io/leaflet-geoman/issues/552


beig:
any updates?

wggley:
Any status?

PR #554, opened 2020-03-06
Update L.PM.Draw.Cut.js

Falke-Design:
Remove second addTo(this._map)


codeofsumit:
Oh my 😵  - was the layer added twice to the map visually as well?

Falke-Design:
No was not added twice, if the layer is already existing on the map / layergroup nothing happening.

PR #557, opened 2020-03-12
Performance 🚀

### LINKED ISSUE #366: Multi-node (6000+) polygon editing performance issue
Hi,

I want to do multi-node (6000+ node) polygon editing. However, the browser runs slowly or crashes because all nodes appear in edit mode.

What can we do to eliminate this problem?

Is it possible to show only the nodes in Bound and edit them in this way? In this way, not all nodes are displayed and no performance problems.

Example jsfiddle: http://jsfiddle.net/wjo5enpq/5/   (Holds a little while loading. Please wait loading.)

I need help,
Thank you.

### LINKED ISSUE #463: How to know when delete is done, or editing is finished?
When a series of `delete` operations are done, finally user clicks `cancel` (or after series of `edits` finally `finish`). From the docs it looks like currently there is no way to identify when the user is `done with deleting` or `done with editing`.

When `onRemove` handlers have high cost functions (such as recalculating all the visible polygons on the screen etc.), it is usually beneficial to batch the operations and do the calculations just once for the whole batch of delete layers. In the Leaflet.draw, the editing and delete operations raise events (e.g. `L.Draw.Event.DELETED`) that inform the list of all layers that were affected for that operation since it began (e.g. `const deletedLayers = event.layers;`)

#### Suggestion: 
Please raise an event on _finish_ of the delete and edit operations, in addition to the existing individual edit and delete events. This gives a chance to group the deleted layers and perform the processing on the group as a batch once the user is done with deleting / editing (and changes the mode to something else)

### LINKED ISSUE #519: Performance issues with geoman and LayersControl
We are loading approx. 4000 CircleMarkers. These are divided into two LayerGroup so we have a basic filter function for these markers. They are loading quite fast, but after we draw a polygon with the geoman library the markers need quite long to render after toggle it within the LayersControl. We also tried it with and without opt-in.

We tried the same feature set with leaflet-draw and hadn't any issues. 

Help would be much appreciated.

codeofsumit:
- [x] Refactoring Modes
- [x] Adding MarkerLimits Mixin
- [x] Adding first option for MarkerLimits Mixin
- [x] Fix the fact that leaflet-geoman makes the app slower just by being there 👎 

This PR aims to fix #519, close #366 and fix #463


codeofsumit:
🚀
perf
perf2
perftest

PR #571, opened 2020-04-05
Swedish translations

TomasJohansson:
There were some misspellings and remaining english words.
Also I have translated "marker" with "markör" instead of "punkt" (similar to the already existing translation of "circle marker" ==> "cirkelmarkör").
However, I can mention that it was intentional to leave the word "punkt" in the following translation:
"finishLine": "Klicka på en existerande punkt för att slutföra",
because I think that it is more appropriate to use the word "punkt" (point) in that sentence.
I am a native swedish speaker, so I can not say the following for sure regarding the english but IMHO I believe that the current english translation should use "point" instead of "marker" here:
"finishLine": "Click any existing marker to finish",
i.e. instead should be:
"finishLine": "Click any existing point to finish",
I think it is normal semantic that a line has many points while a "marker" rather represents some more specific point i.e. not all points in a line should be thought of as markers. IMHO.

### DIFF
--- a/src/assets/translations/sv.json
+++ b/src/assets/translations/sv.json
@@ -18,7 +18,7 @@
   "buttonTitles": {
     "drawMarkerButton": "Rita Markör",
     "drawPolyButton": "Rita Polygoner",
-    "drawLineButton": "Rita Polylinje",
+    "drawLineButton": "Rita Linje",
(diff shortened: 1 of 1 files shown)


codeofsumit:
Thank @TomasJohansson for the detailed explanation and the update 👍 .
I will consider changing the english translation in a future update!

PR #579, opened 2020-04-13
Greek translation

eliotis:


codeofsumit:
Merged it through #611

PR #587, opened 2020-05-12
Add snapping on circle border and L.PM.Utils to global

Falke-Design:
Now layers can be snapped to the circle border.

test

Also added the Utils to the global variable `L.PM.Utils`


codeofsumit:
Wow that's awesome 🎉

Issue #512, opened 2019-10-22
How drag mode works with markers?


NtsDK:
Hi!

I need to listen dragend event on marker. In my expectations I initialized it right but it doesn't work.
I found similar problem here https://github.com/geoman-io/leaflet-geoman/issues/190
And conclusion was listen to pm:edit event.

How drag events should work? Is dragging marker a use case for drag events?

Demo: https://jsfiddle.net/dvh1r2je/

rcollette:
Oddly enough, if you enable a layer with draggable, you do get the drag events.

https://jsfiddle.net/cmbopjae/

```JavaScript
    const marker = L.circleMarker([39.74739, -105]).addTo(map);
    
    marker.on('pm:edit', () => console.log('pm:edit'));
    marker.on('pm:drag', () => console.log('pm:drag'));
    marker.on('pm:dragstart', () => console.log('pm:dragstart'));
    marker.on('pm:dragend', () => console.log('pm:dragend'));
    marker.pm.enable({draggable:true});
```

I was looking at this because I am in some sort of other pickle where my draggable circle marker, basically created the same way as this, doesn't drag but rather the entire map pans instead.  Only thing I can think of is that I am removing a layer from the map prior to this, and there seems to be some known issues with that.

AdamJamesNaylor:
I'm seeing the same issue, I can't seem to bind to any of the drag events for markers.

codeofsumit:
hmm seems like a bug. Thanks for reporting I will take a closer look!

hoangeiu:
Has this problem been fixed?

codeofsumit:
Fixed by #591 

Will be in the next release @hoangeiu

cswww:
Is it solved now? I tried or couldn't

Falke-Design:
Yes this is solved look into: https://jsfiddle.net/dvh1r2je/

PR #591, opened 2020-05-16
Drag events for Marker

Falke-Design:
Adding drag events to marker and circle while edit mode.

Also fixed that CircleMarker fired "pm:edit" twice:
1. Dragging.js -> _dragMixinOnMouseUp -> _fireEdit 
2. _dragMixinOnMouseUp -> "pm:dragend" -> _onMarkerDragEnd -> _fireEdit

Fixed Issue #512


StripedBanana:
I do confirm that regular leaflet events work on simple L.marker objects. Although it seems kind of weird to me that features coming from the geoman sidebar would not be treated the same way regarding events : when I'm creating a marker, I would use regular leaflet events, but when I'm creating anything else I must use geoman `pm:*` events.

So I think it's not really a functionality issue, but more like a conception choice. Thank you for your work on this plugin by the way guys, it's really great 👍

Falke-Design:
Yes @StripedBanana is right. It has only the effect that all layers can use the same event. And the User don't need to think of if it is a Marker or not.

With that it is easier to do following:
```
map.on('pm:create',(e)=>{
  e.layer.on('pm:drag',logEvent)
});
```

instead of:
```
map.on('pm:create',(e)=>{
  if(e.layer instanceof L.Marker){
    e.layer.on('drag',logEvent)
  }else{
    e.layer.on('pm:drag',logEvent)
  }
});
```

StripedBanana:
Actually I was partially wrong. Look at this fiddle I made : https://jsfiddle.net/7r4sfnu5/

When you enable the Geoman's dragging mode, the `dragstart` does not get called anymore on the `L.Marker` (`dragend` still does for some reason). This is problematic to me, as I have no way to use `dragstart` callbacks on my `L.Marker`s when I'm in Geoman's dragging mode.

codeofsumit:
You are right - I now understand the need for this 👍 .

Falke-Design:
@StripedBanana "your" problem was because of the line `this._layer.off('dragstart', this._onPinnedMarkerDragStart, this);` in the `disable` Marker function. This is a pro feature and referenced to it, but it doesn't exists.

I removed it and now it works well.

Falke-Design:
@codeofsumit About what tests you are thinking of? Until now, there are nowhere `drag` events tested. If it should be have tests, I need the PR #592 merged. It adds support for drag in cypress tests.
Readme doesn't need to be updated, because it is already described. There are no excludes for markers.
Merge conflicts are resolved.

Issue #525, opened 2019-12-09
Expand on sample documentation


jasonwinget:
Please add additional examples of how to deal with the data generated by events. Expand on an example of how to use data of creating a line, etc.  I normally understand the 'e' but I am missing the relevant data generated in the event, where is it?  How does someone take this utility to the next level? How do you find your data for storing in a database (not how to store it, but where is the data in the event so a function can be written to do that) for actual real-live use?

map.on('pm:drawend', e => {
  storeMyNewDrawingFunction(e.Something1, e.Something2, etc.);
});

codeofsumit:
Thanks for the feedback. I will improve the documentation on the event payload 👍

Issue #574, opened 2020-04-08
There are no build instructions


sorpigal:
I am not a nodejs sort of person, but I can read instructions and follow them. I have been unable to build geoman myself. There are no instructions so I have done some research on how nodejs and webpack stuff works in general, made some guesses, and ended with a mess.

Can someone who knows how this works produce a step by step guide starting with 'git clone' and ending with a newly-built leaflet-geoman.js?

sorpigal:
Aha, it's in CONTRIBUTING. This is not very clear, but maybe I should have read all .md files just in case.

codeofsumit:
Hi @sorpigal, the  CONTRIBUTING.md is Standard on Github for this but I will make sure to link to it in the README file so people have it easier. Thanks for the suggestion 👍

Falke-Design:
It may be helpful to add the npm scripts as well:
```
npm start --> run webpack dev config with watch mode
npm prepare --> run webpack for release
...
```

swapnasheela12:
hi codeofsumit,

i am trying to use your plugin in angular 9 with ngx-leaflate.there show PM object in MAP object but still got this error

Property 'pm' does not exist on type 'Map'.
please guide me for this.

thanks

Issue #79, opened 2016-09-09
After unsnap, marker of another polygon follows dragged marker


codeofsumit:
![image](https://cloud.githubusercontent.com/assets/2399810/18388462/0990c856-76a2-11e6-828a-ff48463bcd14.png)

wtf? Fix it!

codeofsumit:
ok this has something to do with the raise of markers on hover.

PR #593, opened 2020-05-18
Updated Readme, added Events to the Doc and did some fixes

Falke-Design:
I added a few new things:

- Events to the Readme
- Changed payload of the events to match together f.e. `layer`, `shape`
- Added event `pm:enable` because `pm:disable` also exists
- Added event `pm:langchange`, fired when language is changed
- Added demo event.html with all available events
- Removed `pm:raiseMarkers`, is a old relict #79 
- Fix inconsistent use of `pm:remove` on a layer 
- Added event `pm:globaldrawmodetoggled`, to have a equal event to the mode events. 
  - New functions `map.pm.Draw.`: `getActiveShape()`, `globalDrawModeEnabled()`
- Added developing steps and updated CONTRIBUTING.md

Future ideas / plans / questions:
- Add also `pm:cutmodetoggled` because it is in the edit/mode section of the controls and not clear why this uses the `pm:drawstart` or the new `pm:globaldrawmodetoggled` event
- Functionnames not equal for the `enbaled` function of the modes:
  - Edit Mode:
    - globalEditEnabled
  - Drag Mode:
    - globalDrag**Mode**Enabled
  - Removal Mode:
    - globalRemovalEnabled
- Why has `pm:markerdragstart` and `pm:markerdragend` event in Edit.Line passed indexPath and other types are not? This is inconsistent.

Fix:
#525 
#574


codeofsumit:
Awesome!

Let's get to your questions:
Yes, I think `pm:cutmodetoggled` makes more sense for cutting. Let's add it but the event itself should still fire `drawstart` for backwards compatibility so we don't introduce a breaking change here.

Regarding the inconsinstent function names: Would love a normalization here. Just make sure they are backwards compatible with a comment to remove them in the next major release.

Re indexPath in `pm:markerdragstart` - because it is not available in other types than L.Edit.Line - in order to find a marker/coord in the coordinate array, a path is needed. This is not needed for Circles or Markers. IMO this can stay like this. What would be your suggestion to change here?

README.md
```diff
@@ -378,17 +411,26 @@ map.pm.toggleGlobalRemovalMode();
...
+| :---------- | :----- | :------------------------------------------------------- | :------ |
+| pm:remove   | `e`    | Fired when a layer is removed via Removal Mode           | `layer` |
 
 The following events are available on a map instance:
 
-| Event       | Params | Description                                              |
-| :---------- | :----- | :------------------------------------------------------- |
-| pm:remove   | `e`    | Fired when a layer is removed via Removal Mode           |
-| layerremove | `e`    | Standard Leaflet event. Fired when any layer is removed. |
+| Event                         | Params | Description                                              | Output            |
```
codeofsumit, line 429: This should theoretically add a fourth column to the table, right? Somehow, VSCode is not showing this to me but I see the code in the README 🤔

image

Does "Output" has another effect that I'm not aware of?

README.md
```diff
@@ -575,6 +625,46 @@ I'm adopting the Issue Management of lodash which means, feature requests get th
 You can upvote existing feature requests (or create new ones). Upvotes make me see how much a feature is requested and prioritize their implementation.
 Please see the existing [Feature Requests here](https://github.com/geoman-io/leaflet-geoman/issues?q=is%3Aissue+is%3Aclosed+label%3A%22feature+request%22+sort%3Areactions-%2B1-desc) and upvote if you want them to be implemented.
 
+### Developing
```
codeofsumit, line 628: 👌

demo/events.js
```diff
@@ -0,0 +1,80 @@
+var map = L.map('map').setView([39.74739, -105], 13);
```
codeofsumit, line 1: I think a cypress test for all events would be great too (at least all events that can be tested)

README.md
Falke-Design: In my editor it looks well and in [stackedit](https://stackedit.io/app#) it looks also good. I don't know what I should change 🤷‍♂️

Output is the payload

Falke-Design:
I removed the eslint errors, sorry I didn't know about this feature.

PR #594, opened 2020-05-18
Marker drag check

### LINKED ISSUE #567: enableDraw causes error if one marker is set to interactive: false
When on or more marker on map are set to not interactive, enableDrawn('Marker') causes error 

TypeError: this._layer.dragging is undefined

See code to reproduce here

https://jsfiddle.net/VoyagesDivins/qe6yxba3/9/

May be linked to  #540 ?

Falke-Design:
Check if marker has dragging property before enable / disable dragging

Fix #567 

When a marker is inactive (interactive: false) nothing can be done with it, but other layers still snapping to it. Maybe snapping should be also removed.


Falke-Design:
@codeofsumit Both things added

Falke-Design:
Yes it is difficult to understand. 
Before this changes, when a interactive Marker was added to map and then edit mode was enabled, it thrown a error because the marker was not draggable.

With this test it only checks if a error is thrown when goes into edit mode.

Falke-Design:
Can a cypress test on github restarted? It stopped before tested ...
Local all tests are passed.

Issue #517, opened 2019-11-07
Geoman optIn


Jumpy555:
Hi,
I'm using Geoman library in a map with thousands of markers.

To display markers on the map I also use leaflet markercluster.

As you can see in this fiddle, if I don't include Geoman library, the map is loading very fast:
https://jsfiddle.net/odr2Lzgh/

When I include Geoman, the map is loading much slower:
https://jsfiddle.net/t09djpsm/

I even tried to add L.PM.initialize({ optIn: true }); to avoid markers being taken into consideration by Geoman, but it did not have any effect.

Is there a way to initialize Geoman and specify to avoid the markers layer?

Thanks, 
Jumpy

Dan-Wood:
Your second fiddle doesn't include Geoman at all, wrong link :) ?

This with 10k works just fine, and fast :)

https://jsfiddle.net/adfvzg7t/1/

Jumpy555:
If you look at the resources, you'll see that in the second fiddle I included Geoman library.

I didn't put the control on the map to show that, just including Geoman as a resource, the same javascript code is much slower (almost 0 seconds vs. 13 seconds).

You would say that, if I don't use it on the map, I should not include it in my resources, but I need to have a shared minified library with Leaflet, Geoman and other libraries.

What I'd like is to have the possibility to initialize Geoman only when needed on the map or, at least, specify to ignore the whole markers layer because I only need the drawing functions.

Jumpy

codeofsumit:
Thanks @Jumpy555 - I'll get on this soon. Let me know if you have more infos, demos, sample data files that could help me figure this out

PR #595, opened 2020-05-18
Add reInitLayer function: Enable leaflet-geoman on an ignored layer:

Falke-Design:
Fix: 
#533 
Maybe also #516 I don't understand it exactly

Issue #569, opened 2020-04-03
Question: [OPT-IN] Newly created shapes


JanLoebel:
Hi Summit,

I know (have read) that you will rework the current "OPT-IN"-mechanism. But currently I'm started to work with it and wonder if there is an easy way to at least "opt-in" the created lines/polygons etc.. Currently I'm doing it like this:

```
map.on('pm:create', (pmCreateEvent: any) => {
      // XXX Workaround for a problem regarding opt-in
      const layer = pmCreateEvent.layer;
      layer.options.pmIgnore = false;
      const shape = pmCreateEvent.shape;

      if (shape === 'Line') {
        layer.pm = new L.PM.Edit.Line(layer);
      } else if (shape === 'Marker') {
        layer.pm = new L.PM.Edit.Marker(layer);
      } else if (shape === 'Rectangle') {
        layer.pm = new L.PM.Edit.Rectangle(layer);
      } else if (shape === 'CircleMarker') {
        layer.pm = new L.PM.Edit.CircleMarker(layer);
      } else if (shape === 'Polygon') {
        layer.pm = new L.PM.Edit.Polygon(layer);
      } else if (shape === 'Circle') {
        layer.pm = new L.PM.Edit.Circle(layer);
      }
}
```

Is there a better way to start with it currently?

Best regards,
Jan

Issue #533, opened 2020-01-20
Change pmIgnore after adding geojson


arshdoda:
Is there any way to modify the pmIgnore after adding the featuregroups.

DazDotOne:
@arshdoda It doesn't seem to be implemented as yet.

I struggled with this quite a bit. I've put an example  of how I worked around it [in an issue I started](https://github.com/geoman-io/leaflet-geoman/issues/516) regarding this feature.

However, I will say that drawing, draging, editing, and deleting are all handled slightly differently.

I'd recommend cloning the repo and looking at how things are done to see what behaviour you need to override.

most of the behaviour is in the L.PM.Map class but could be anywhere.

PR #596, opened 2020-05-19
Re-Designed the optIn

Falke-Design:
I thought about a new design for the optIn. 

**Current behaviour:**
1.  `L.PM.initialize();` is called
2. Hooks are added to the Classes: `L.Polyline.addInitHook(initPolyline);`
3. `L.PM.initialize({optIn: true);` is called
4. Hooks are added second time to the Classes: `L.Polyline.addInitHook(initPolyline);` with `optIn: ture`

Then creating layer:
1. Layer created
2. First initHook is called
3. Set `layer.pm = undefined` 
4. `layer.pm` is created if `pmIgnore` is not `true`
5. Second initHook is called
6. Set `layer.pm = undefined` 
7. `layer.pm` is created if `pmIgnore` is `false` (because `optIn = true`)

With that behaviour the `layer.pm` object is created and in the second call the object is deleted if `optIn` is set (if `pmIgnore` is not `false`).

**New design:**
1.  `L.PM.initialize();` is called
2. Hooks are added to the Classes: `L.Polyline.addInitHook(initPolyline);`
3. `L.PM.setOptIn(true);` is called

Then creating layer:
1. Layer created
2. initHook is called
3. Set `layer.pm = undefined` 
4. `layer.pm` is created if  `L.PM.optIn` is `true` and  `pmIgnore` is `false` **OR** `pmIgnore` is not `true`

With that behaviour the `layer.pm` object is only created if it is allowed and initHook is not called twice.
And the `optIn` can be disabled: `L.PM.setOptIn(false)`

Fix:
#517 
#533
And has also included function from PR #595:
- #569
- Maybe also #516 I don't understand it exactly


Falke-Design:
@codeofsumit tests added

codeofsumit (approved):
This will make many people happy 👏

Issue #576, opened 2020-04-09
PreventMarkerRemoval prevents removal of layers after going into edit mode


schoolmeister:
### Bug
Hey, I think I've found a bug with `preventMarkerRemoval` of which I'm unsure whether it is a bug or whether it is supposed to be like this. From my understanding from the docs, `preventMarkerRemoval` only prevents removal of a layer by right-clicking, but removal using the removal tool is still possible. Yet, when it is enabled I cannot remove layers after going into editing mode.
### Reproduce
1. Enable `preventMarkerRemoval` (either globally or on a single layer)
2. Create a new layer
3. Go into edit mode
4. Go into removal mode and try to remove the layer.
5. The layer cannot be removed anymore. 

When skipping the edit mode step removal is still possible.

As a quick example, here's a JSFiddle that goes into edit mode right after creating a layer: https://jsfiddle.net/gtvbzj1d/1/

PR #597, opened 2020-05-19
Fix Bug: preventMarkerRremoval

Falke-Design:
Added check for Marker or CircleMarker to the function:

```
enableGlobalRemovalMode() {
    const isRelevant = layer =>
      layer.pm &&
      !((layer instanceof L.Marker || layer instanceof L.CircleMarker) && layer.pm.options && layer.pm.options.preventMarkerRemoval) &&
      !(layer instanceof L.LayerGroup);
```

Fix: #576


src/js/Mixins/Modes/Mode.Removal.js
```diff
@@ -16,7 +16,7 @@ const GlobalRemovalMode = {
   enableGlobalRemovalMode() {
     const isRelevant = layer =>
       layer.pm &&
-      !(layer.pm.options && layer.pm.options.preventMarkerRemoval) &&
+      !((layer instanceof L.Marker || layer instanceof L.CircleMarker) && layer.pm.options && layer.pm.options.preventMarkerRemoval) &&
```
codeofsumit, line 19: Can you add a comment here explaining when a layer is relevant?
I would also like to see a test for this - otherwise it will be quite hard to debug when this causes a bug or unexpected behaviour in the future

Falke-Design: I re-thinked about that ... Why is it checking for `preventMarkerRemoval`? Because it only adds a `click` event and that is allowed for all Markers no matter if the option `preventMarkerRemoval` is passed. 

`preventMarkerRemoval` only restrict that a Marker can't deleted by right clicking (contextmenu).

My suggestion is to remove the complete line.

codeofsumit: You mean the `preventMarkerRemoval` option has no effect here right now? I think you are right, in which case this can be deleted.

Falke-Design:
@codeofsumit I deleted the line with `preventMarkerRemoval`

PR #598, opened 2020-05-19
Cut: Prevent that a unnecessary layergroup is created

### LINKED ISSUE #572: Cut creates ghost layers that stays permanently on leaflet layers array
First, thanks for your amazing library! Really helpfull!

Repro https://jsfiddle.net/pj79x3vm:

- Draw two rectangles:
image

    - Console shows 2 layers added from draw functions: 2 features with `Polygon` type.

- With a single cut, divide them in half resulting 4 rectangles:
image

    - Console shows 4 layers added from draw functions: 2 features with `MultiPolygon` type and 2 feature collections with a single feature each of type `MultiPolygon`.

- Use the removal tool to clear all polygons;

- Check the console: we still have 2 layers as `"FeatureCollection"` returning from `map.eachLayer(layer => { ... })`

Is this the expected behaviour? I guess i can filter by `"Feature"` only.

Falke-Design:
Fix #572 

The problem was that `L.geoJSON` always creating a layergroup, also if it has only one layer.


Falke-Design:
I have no idea how to test something so small and unimportant 😅

AdamJamesNaylor:
@Falke-Design if you take a look inside `cypress\integration\globalmodes.spec.js` you should be able to copy one of those tests and use something like `cy.hasLayers(x)` to test the correct number of layers are being created?

This issue is also causing issues in my project, again relating to when `toGeoJSON()` is called 👍

Falke-Design:
@codeofsumit test added

codeofsumit:
@AdamJamesNaylor that's exactly how to test it 👍 
@Falke-Design will go through it Monday and merge it asap

Great work 👏

PR #562, opened 2020-03-20
LayerGroup Refactoring

codeofsumit:
LayerGroup support is flaky at the moment. A refactoring is needed and aims to fix #561

PR #599, opened 2020-05-19
Bug: enable edit for new added layers

### LINKED ISSUE #561: Features added to LayerGroups are not properly enabled when already in Global Edit Mode
When adding features mid-edit to a group that is already on the map

```

map.pm.enableGlobalEditMode();


addFeature(feature1);
addFeature(feature2);
addFeature(feature3);


function addFeature(feature) {
  
  var group = L.geoJson({
    "type": "FeatureCollection",
    "features": []
  }).addTo(map);
  
  L.geoJson(feature).addTo(group);
}

```

Fiddle:
https://jsfiddle.net/bLnx5e2g/

Falke-Design:
Removed throttle because event must fired for each new layer so it can be enabled.

With the throttle the event was called only every 100ms and not for each layer.

Fix #561 
Your PR: #562


codeofsumit:
Great find of the root cause 👍. Unfortunately, this will significantly reduce performance. Let's try to find a solution that keeps the throttle, yet fixes #561.

PR #601, opened 2020-05-23
Added css for toolbar / controls on the right side

Falke-Design:
Old:

New:

Sorry for the bad quality  ^^


codeofsumit:
Can you please provide a description and screenshot of the problem? It's hard to tell what is being fixed/improved here when just looking at the CSS 🙂

Falke-Design:
@codeofsumit added

PR #603, opened 2020-05-31
Make CircleMarkers editable and updating snapping to circle borders

Falke-Design:
When global option `editable` is set to `true` a CircleMarker can be drawn and edited like a normal Circle.

Also added a update for snapping to circle borders. => If this PR will not be merged, I create a new PR with this update.


codeofsumit:
Snapping to Circle Borders is pretty cool 😍 .
What I don't understand is why a CircleMarker needs to be editable like a circle? Can you show a GIF of that?
If it's exactly like a circle, why don't just use circle? What is the difference then?

Falke-Design:
There are the same differences with edit and without 😄 A circlemarker has a fixed pixel size which has always the sime size at each zoom level, a circle is resized when zooming. 

I thought everything is editable, only the circlemarker is not. 
This allows the user to change the size of a circlemarker and makes it more customizable.

codeofsumit:
ahhh got it, nice!
After merge conflicts are resolved, I will merge 👍

PR #609, opened 2020-06-21
Adds support for merging polygons

AdamJamesNaylor:
Adds a new toolbar button that activates merge mode. You can then click two polygons to perform a union.

Raises a pm:merge event which includes the two participating layers and the resulting union.

The button can be excluded from the toolbar by passing { mergeMode: false } to addControls().

Introduces a dependency on @turf/union version ^6.0.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,11 @@
 
 ![Demo](https://file-gmeileqfmg.now.sh/)
 
+### Using Leaflet-Geoman in production?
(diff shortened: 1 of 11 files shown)


AdamJamesNaylor:
Need to amend this to remove `pm:remove` as the documentation says this should only be called when a layer is removed via removal mode.

Issue #608, opened 2020-06-19
Ability to union two or more polygons into one


AdamJamesNaylor:
One of my use cases is the ability for a user to draw two overlapping polygons that can then be combined/merged into a single polygon.

Likely using something like;
`var union = turf.union(poly1, poly2);`

The [documentation](http://turfjs.org/docs/#union) shows an example.

Will need a discussion around how the UX will be handled.

My initial proposal is to;

1. First add the ability for a user to select multiple polygons at once (including a TBD indication of that selection), possibly using the `ctrl` key.
2. Then add a toolbar that appears near/next to the selection including contextual buttons. For now just `union`.
3. Once clicked, the overlap is removed and the resulting single union remains.
4. If no overlap exists nothing happens.
5. No pre-checks are performed (like only showing the button if an overlap exists) as the number and complexity of the polygons selected could be significant.

If this sounds adequate @codeofsumit, I'll get a fork up together.

AdamJamesNaylor:
Initial PR [here](https://github.com/geoman-io/leaflet-geoman/pull/609).

Falke-Design:
Closed and will be implemented if it has enough upvotes

PR #610, opened 2020-06-21
Adds support for merge mode.

AdamJamesNaylor:
Adds a new toolbar button that activates merge mode. You can then click two polygons to perform a union.

Raises a `pm:merge` event which includes the two participating layers and the resulting union.

The button can be excluded from the toolbar by passing `{ mergeMode: false }` to `addControls()`.

Introduces a dependency on `@turf/union` version ^6.0.3.

### DIFF
--- a/cypress/integration/globalmodes.spec.js
+++ b/cypress/integration/globalmodes.spec.js
@@ -1,5 +1,41 @@
 describe('Modes', () => {
   const mapSelector = '#map';
+  it('unions two overlapping polygons into one polygon', () => {
+
+    cy.toolbarButton('polygon').click();
+
+    cy.get(mapSelector)
+      .click(100, 100)
+      .click(300, 100)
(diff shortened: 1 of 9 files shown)


AdamJamesNaylor:
Addresses issue #608

codeofsumit:
Hi @AdamJamesNaylor 

Thanks a lot for your work on this!
A merge feature itself is great. I do have some bigger plans for this though which includes selection and batch operations like delete, merge, subtract, drag, etc.
Which means unfortunately I can't merge this as I need to build this with a large architectural change.

AdamJamesNaylor:
Ok thanks @codeofsumit 
Do you have a rough timescale on when that work is likely to be started or merged into master? As it's something my project will make heavy use of. Including some of the other features you mentioned.

Would you be open to work collaboratively on it? Would love to help speed it up 👍

Issue #331, opened 2018-08-14
Invalid geometry on double click


ka7eh:
If I activate draw mode with `selfIntersection` set to `false`, then double clicking on the map over the last added vertex creates a new point which produces a self-intersecting shape. After that, you cannot finish your geometry, because it is not valid.

See this fiddle for an example: https://jsfiddle.net/ny7jbv5t/9/
Start drawing a polygon and at one point double click on your last vertex. Then try to finish your edit, which should fail.

codeofsumit:
@ka7eh double click will create two vertexes on top of each other. What would you expect on double click to happen?

ka7eh:
When `selfIntersection` is set to `false`, it should not create the second vertex, because it's an invalid geometry.

codeofsumit:
true, I will check this

codeofsumit:
Does anyone know the definition of a self-intersection?

Because it seems like this:
1. click, unique coords: selfIntersection = `false`.
2. click, unique coords: `false`
3. click, same coords as 2nd click: `false` <-- 🤔
4. click, unique coords: `true` <-- too late 🤔🤔

I would've expected the same coord is already invalid. So it should stop the 3rd click already 🤔🤔. Does anyone have more info?

tim-salabim:
Maybe it cannot be detected at the time of adding the vertex, because a duplicated vertex as such is not a violation but rather a necessity to produce a valid polygon (1st and last vertex identical)?

codeofsumit:
@tim-salabim good thought. Afaik those identical vertices need to be at the first and last index of the coords array while in this scenario, they're the last and second to last 🤔  It's so weird.

Falke-Design:
@codeofsumit why you not check if latlng from the last (first-click) vertex is equal to the second-click vertex?

In Draw.Line._createVertex():
```
if(!this.options.allowSelfIntersection 
    && latlng.equals(this._layer.getLatLngs()[this._layer.getLatLngs().length-1])){
      return;
}
```

Falke-Design:
Fixed in Release:  `2.7.0`

PR #621, opened 2020-07-02
Fix duplicate point creation in polygon draw

m-hogg:
Double clicking during polygon creation would create two identical markers. This would cause the polygon drawing to break when self intersection was not allowed, as users could not place another marker anywhere.

See issue #331


codeofsumit:
Nifty solution, I like it!
Would like @Falke-Design's opinion as well before I merge it

Falke-Design:
Fix #631

PR #626, opened 2020-07-07
Toolbar block position

Falke-Design:
Add function `map.pm.Toolbar.setBlockPosition(block,position)` to set position of each block

grafik

### DIFF
--- a/README.md
+++ b/README.md
@@ -667,7 +667,7 @@ map.pm.Toolbar.getControlOrder();
 
 **Position**
 
-You can set different positions per block `draw`, `edit`, `options`⭐, `custom`, `oneBlock`: 
+You can set different positions per block `draw`, `edit`, `options`⭐, `custom`: 
 
 Possible values are `'topleft'`, `'topright'`, `'bottomleft'`, `'bottomright'`
 
--- a/README.md
+++ b/README.md
@@ -657,13 +657,38 @@ map.pm.addControls({
 
 Reorder the buttons with
 ```js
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -657,14 +657,39 @@ map.pm.addControls({
...
 
 Receive the current order with
 ```js
-map.pm.Toolbar.getControlOrder()
+map.pm.Toolbar.getControlOrder();
 ```
 
+**Position**
+
+You can set the position of each block:
```
codeofsumit, line 670: let's change this to 
`You can set different positions per block`

README.md
```diff
@@ -657,14 +657,39 @@ map.pm.addControls({
...
+
+```javascript
+map.pm.addControls({
+  positions: {
+    draw: 'topright',
+    edit: 'topleft',
+    options: undefined,
+    custom: undefined,
+    oneBlock: undefined,
```
codeofsumit, line 679: let's not encourage to pass `undefined` here. Remove the option completely from the example or pass `false`.

src/js/Toolbar/L.PM.Toolbar.js
```diff
@@ -23,6 +23,13 @@ const Toolbar = L.Class.extend({
     customControls: true,
     oneBlock: false,
     position: 'topleft',
+    positions: {
+      draw: undefined,
```
codeofsumit, line 27: same here. `undefined` should not be the default. `false` should.

codeofsumit (changes requested):
That's amazing! Some minor things but then we can merge it.

- [ ] See line comments
- [ ] A test 😁

PR #627, opened 2020-07-09
Fix bug when circle is the only layer on the map

Falke-Design:
When circle was the only layer on the map, the hidden snapping-border circle was making problems while edit mode.

Fix: #640
Fix: #652

### DIFF
--- a/src/js/Edit/L.PM.Edit.Circle.js
+++ b/src/js/Edit/L.PM.Edit.Circle.js
@@ -6,6 +6,8 @@ Edit.Circle = Edit.extend({
   initialize(layer) {
     this._layer = layer;
     this._enabled = false;
+    // create polygon around the circle border
+    this._updateHiddenPolyCircle();
   },
   applyOptions() {
     if (this.options.snappable) {
--- a/src/js/Edit/L.PM.Edit.CircleMarker.js
+++ b/src/js/Edit/L.PM.Edit.CircleMarker.js
@@ -6,6 +6,8 @@ Edit.CircleMarker = Edit.extend({
   initialize(layer) {
     this._layer = layer;
     this._enabled = false;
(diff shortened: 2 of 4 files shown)


codeofsumit:
This seems to be quite custom.
Couldn’t we utilize `pmTempLayer` For this?

https://github.com/Falke-Design/leaflet.pm/blob/circle_bug/src/js/Mixins/Snapping.js#L264

Maybe I misunderstand the problem here - can you explain it more?

Falke-Design:
No because of the line from you, the snapping will not work on `_pmTempLayer` layers. So I can't use this flag for the hidden Circle, where I want to snap.

So when I have one Circle on the map and I edit them, then a hidden Circle for snapping to the borders is created. Because of this new hidden layer, the available snapping layers is greater then 1. Then a snapList will be created. 
But in the function `_calcClosestLayer` the snapping to the own circle border from the "parent" circle is not allowed [Snapping.js#L282](https://github.com/Falke-Design/leaflet.pm/blob/8df1b149e66e55aa76d0837cda9cf845f01b4d2f/src/js/Mixins/Snapping.js#L282). Now no other layer is on the map where can be snapped, so [Snapping.js#L277](https://github.com/Falke-Design/leaflet.pm/blob/8df1b149e66e55aa76d0837cda9cf845f01b4d2f/src/js/Mixins/Snapping.js#L277) will returned empty and then it runs in `_checkPrioritiySnapping()` into a empty object and crashs [Snapping.js#L171](https://github.com/Falke-Design/leaflet.pm/blob/8df1b149e66e55aa76d0837cda9cf845f01b4d2f/src/js/Mixins/Snapping.js#L171).

When more then one Circle is drawn on the map, there will never returned an empty Object.

PR #628, opened 2020-07-09
Layers can now directly added to layergroups

Falke-Design:
Layers can now directly added to layergroups while creation.

Example:
```
const fg = L.featureGroup().addTo(map);
map.pm.setGlobalOptions({layerGroup: fg});
```
Now the drawn layers added to the featuregroup instead to the map.


src/js/L.PM.Map.js
```diff
@@ -131,6 +132,9 @@ const Map = L.Class.extend({
   },
   getGeomanDrawLayers(){
     return findLayers(this.map).filter(l => l._drawnByGeoman === true);
+  },
+  _getLayerGroup(){
+    return this.globalOptions.layerGroup && this.globalOptions.layerGroup instanceof L.LayerGroup ? this.globalOptions.layerGroup : this.map;
```
codeofsumit, line 137: Mhh.. a function named `getLayerGroup` should not return the map instance by default.

Please rename to `_getContainingLayer()` and document in code & README with something like 

`// returns the map instance by default or a layergroup is set through global options`

codeofsumit (changes requested):

src/js/L.PM.Map.js
Falke-Design: Good point! I changed it but I didn't described it in the Readme because it is a "private" function

codeofsumit (approved):

Issue #631, opened 2020-07-15
doubleclick on map get's polygon stuck in "self intersection" state


smeijer:
When using both `allowSelfIntersection: false`, and `finishOn: dblclick` to finish polygons, the double click gets the polygon stuck in an `intersecting` state.

To reproduce: [jsfiddle](https://jsfiddle.net/smeijer/4x8or2n5/3/)

1. Initialize with:

```js
map.pm.addControls({
  position: 'topleft',
});

map.pm.setGlobalOptions({
  allowSelfIntersection: false,
  finishOn: 'dblclick',
});
```

2. Click the `draw polygon` button

3. Create a polygon with 3 (or more) vertices

4. Finish by clicking the mouse button twice (`dblclick`).

5. Polygon is now stuck in an `intersecting` state, so it can't be closed/finished.

![pm-self-intersect](https://user-images.githubusercontent.com/1196524/87561353-da7d0880-c6bc-11ea-884f-f4a9dfd1feca.gif)

smeijer:
I've updated the title, as it turns out that it doesn't have anything to do with the `finishOn: 'dblclick'` part. Also without that setting, the polygon gets stuck when double-clicking the map.

Falke-Design:
Thx for reporting! This is already fixed in the develop branche https://github.com/geoman-io/leaflet-geoman/pull/621
But I found a new bug in the develop version. When dblclick is called no vertex is created.

smeijer:
That's awesome! 

Regarding your finding. Is that a bug, or a feature? I think it is actually a nice user experience if a double click doesn't add a vertex. Especially in the mode where double click finishes the feature. This would prevent accidental vertex insertions, where users don't won't to add a vertex, but just want to finish the feature.

Falke-Design:
We can discuss about this but I use it to create the last vertex and then finish it at the same time. When I want to finish the polygon without creating a vertex I click on the first marker.

@codeofsumit how do you see that?

smeijer:
Would it be possible to publish a patch or prerelease that contains this bug fix?

codeofsumit:
I will release a patch version within the next few days 👌

JamboBuenna:
@codeofsumit Any idea when this patch is going to go out as an npm package. 
Picked up the code change on develop and this fix works great, but having problems getting our build pipeline working against your dev branch and would prefer to fix to an npm release.

codeofsumit:
[...]

PR #632, opened 2020-07-15
Fix finishOn dblclick bug

Falke-Design:
Because of the PR https://github.com/geoman-io/leaflet-geoman/pull/621 no vertex is created when finished with `dblclick`.

Reference #631


codeofsumit:
Love that you've added a test to check on this case 👏👏👏

Issue #630, opened 2020-07-14
Cut bug


Armaggik:
![Geoman](https://user-images.githubusercontent.com/68292888/87450814-35fdb680-c629-11ea-9cbc-42c1261b37a8.gif)

Falke-Design:
@codeofsumit This is happening because of to much multi-polygon holes. When I flatten the multi-polygon to single layers it works

codeofsumit:
I'm not aware of a limit in MultiPolygons. Is that a bug in Turf?

Falke-Design:
Test what should be passed:

```
  it('test cut bug', ()=>{
    cy.toolbarButton('rectangle').click();
    cy.get(mapSelector)
      .click(191,216)
      .click(608,323);

    cy.toolbarButton('cut').click();
    cy.get(mapSelector)
      .click(226,389)
      .click(230,105)
      .click(270,396)
      .click(226,389);

    cy.toolbarButton('cut').click();
    cy.get(mapSelector)
      .click(293,356)
      .click(293,122)
      .click(340,367)
      .click(293,356);

    cy.toolbarButton('cut').click();
    cy.get(mapSelector)
      .click(364,345)
      .click(363,138)
      .click(414,368)
      .click(364,345);

    cy.toolbarButton('edit').click();
    cy.hasVertexMarkers(16);
  })
```

@codeofsumit yes it is because of Turf.js (https://github.com/Turfjs/turf/issues/1922)

Falke-Design:
Fixed in Release:  `2.7.0`

Issue #555, opened 2020-03-09
Strange behavior while cutting 


Mixser:
Hello, I faced with strange behavior while cutting. I have used the [geoman-editor](https://geoman.io/geojson-editor). I draw a polygon and want to cut a hole (including the one or more edges of polygon). And sometimes the edge is disappearing sometimes not and this is strange. Maybe this behavior relay to issue #396, but i'm not sure.

Expected behavior: the edge should disappear while cutting.

![geoman_bug](https://user-images.githubusercontent.com/1364297/76219093-51ddef80-6226-11ea-8265-c69c0d312ed7.gif)

Falke-Design:
This is a bug of turfjs resp. the library [martinez](https://github.com/w8r/martinez) used from turfjs.
I don't know how to describe that bug, so pls follow the steps.
Look into the [jsfiddle](https://jsfiddle.net/falkedesign/4nhLpb3r/). **Zoom lvl 27!**
1. Click on the Cut button --> The Polygon is nearly cutted correct. (It's not cutted on the correct horizontal line, but has the hole)
2. Move the blue polygon over the line from the red polygon
3. Click on the Cut button --> The Polygon is not cutted correct.

I hope you understand what I mean.

I have no idea how to solve this problem. There are a few [issues ](https://github.com/w8r/martinez/issues)with the difference function and also a PR but I don't know if this helps us.

PR #633, opened 2020-07-15
Replace difference and intersection from turf

Falke-Design:
Replace difference and intersection from turf with the new lib `polygon-clipping`. This lib is also used in the alpha version of turf but turf make problems ... so why not use the original source?

Fix: #630

### DIFF
--- a/cypress/integration/rectangle.spec.js
+++ b/cypress/integration/rectangle.spec.js
@@ -67,4 +67,35 @@ describe('Draw Rectangle', () => {
       expect(marker.options.draggable).to.equal(false);
     });
   });
+
+  it('Multiple Cuts', ()=>{
+    cy.toolbarButton('rectangle').click();
+    cy.get(mapSelector)
+      .click(191,216)
+      .click(608,323);
+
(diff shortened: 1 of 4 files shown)


codeofsumit:
I like replacing turf where possible.
Can you give a bit more info on:

Library size: how's the delta between the components we removed vs what we import now?

Functionality:
Anything we have to be aware of that the new library is missing or edge cases which are unsupported?

Falke-Design:
Library size:
https://bundlephobia.com/result?p=@turf/difference@6.0.2 8.2kB
https://bundlephobia.com/result?p=@turf/intersect@6.1.3 6.8kB
replaced by
https://bundlephobia.com/result?p=polygon-clipping@0.15.1 7.7kB

Functionality:
Only difference and intersection is replaced what currently only used in `Draw.Cut`. While testing by my self and with cypress no problems happen and I found no unsupported cases.
Also I tried #555 but this is still a issue.

Also a positiv fact is that the library is up-to-date. The latest realease is 8 days old.
https://github.com/mfogel/polygon-clipping

Issue #413, opened 2019-03-20
[Question] How to trigger cancel automatically


pichaya07:
Hi!

I would like to trigger a "cancel" event.
How can I trigger this event?

![Screen Shot 2562-03-20 at 11 33 56](https://user-images.githubusercontent.com/8343315/54660632-bbdb2180-4b09-11e9-81a5-054c54d42fca.png)

Falke-Design:
with
```map.pm.enableDraw('Line');
    map.pm.disableDraw('Line');
```
you can force a cancle, but i think the event 'cancle' is not fired

pate313373:
Very nice project! I really like it! :)

I have a HTML select element where the user should select the layerGroup he wants to draw in. If none is selected, I want the draw/edit operation to be cancled.

I tried as suggested:
`mymap.on('pm:drawstart', ({ workingLayer }) => {
					if(typeof userLayers[$('#userLayers_sel').val()] === 'undefined'){
						mymap.pm.enableDraw('Line');
						mymap.pm.disableDraw('Line');
						alert('No layer selected!');
					}
					workingLayer.on('pm:vertexadded', e => {
						console.log(e);
					});
				});`

However, it seems the suggested code fails with "Uncaught RangeError: Maximum call stack size exceeded".

Is there any other way to fire a cancle directly? In my opinion it would be a way cleaner approach.

Thanks and best regards

Falke-Design:
Hi @pate313373 you run into this error, because you create a loop. When you call `map.pm.enableDraw('Line')` the `pm:drawstart` event is called, and in the event listener you call again `enableDraw`.

The better version is:
```
map.on('pm:drawstart',(e)=>{
	const shape = map.pm.Draw.getActiveShape();
	if(shape){
		map.pm.disableDraw(shape)
	}
})
```

**BUT** we have a bug in the Library so I have to fix this first. The `pm:drawstart` is called to early.

Until them you have to use a workaround with `setTimeout`:

```
map.on('pm:drawstart',(e)=>{
	setTimeout(()=>{
		const shape = map.pm.Draw.getActiveShape();
		if(shape){
			map.pm.disableDraw(shape)
		}
	},0.5);
})
```

pate313373:
Excellent, thx  for the fast help! :)

PR #639, opened 2020-08-07
Restructure

Falke-Design:
Restructure files:
- Change point of fire enable/disable events --> Moved to the end of the functions
- Changed a lot of file, so that they have the same structure --> enable, disable, enabled, toggle, ... create ... move ... remove. Also moved functions below they function from where it called (snapping)

Fix: #413

### DIFF
--- a/src/js/Edit/L.PM.Edit.CircleMarker.js
+++ b/src/js/Edit/L.PM.Edit.CircleMarker.js
@@ -7,54 +7,6 @@ Edit.CircleMarker = Edit.extend({
     this._layer = layer;
     this._enabled = false;
   },
-  applyOptions() {
-    // Use the not editable and only draggable version
-    if (!this.options.editable && this.options.draggable) {
-      this.enableLayerDrag();
-    } else {
-      this.disableLayerDrag();
-    }
-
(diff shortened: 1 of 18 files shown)


src/js/Edit/L.PM.Edit.CircleMarker.js
```diff
@@ -104,18 +105,17 @@ Edit.CircleMarker = Edit.extend({
     // disable dragging, as this could have been active even without being enabled
     this.disableLayerDrag();
 
+    this._layer.off('contextmenu', this._removeMarker, this);
+
     // only fire events if it was enabled before
     if (!this.enabled()) {
```
Falke-Design, line 111: @codeofsumit 
`CircleMarker` is the only Shape which checks if it was enabled to fire events. Should we remove this or add this to the other Shapes too?

codeofsumit: remove it for now and lets see if it creates problems. I guess there was a reason for this but my comment should have stated it

codeofsumit (changes requested):
Will merge after merge-conflicts are resolved

PR #647, opened 2020-08-21
Add Hungarian translation

szimih:


codeofsumit:
Awesome @szimih 👏  - thanks for contributing!

PR #649, opened 2020-08-27
Add toGeoJSON() to getGeomanLayers() and getGeomanDrawLayers()

Falke-Design:
Add `toGeoJSON()` to `getGeomanLayers()` and `getGeomanDrawLayers()`

Example:
```
map.pm.getGeomanLayers().toGeoJSON()
```

I add the function `toGeoJSON()` to the layers result array, so that it is possible to call a function "after" a function. Are there problems when we do this?

### DIFF
--- a/src/js/L.PM.Map.js
+++ b/src/js/L.PM.Map.js
@@ -126,27 +126,29 @@ const Map = L.Class.extend({
   disableGlobalCutMode() {
     return this.Draw.Cut.disable();
   },
-  getGeomanLayers(){
+  getGeomanLayers(asGroup = false){
     const layers = findLayers(this.map);
-    layers.toGeoJSON = function(num){ // use function because this have to be the layers array
-      const group = L.featureGroup();
-      this.forEach((layer)=>{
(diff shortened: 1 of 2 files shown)


src/js/L.PM.Map.js
```diff
@@ -127,12 +127,27 @@ const Map = L.Class.extend({
...
-    return findLayers(this.map);
+    const layers = findLayers(this.map);
+    layers.toGeoJSON = function(num){ // use function because this have to be the layers array
+      const group = L.featureGroup();
+      this.forEach((layer)=>{
+        group.addLayer(layer);
+      });
+      return group.toGeoJSON(num);
+    };
+    return layers;
```
codeofsumit, line 138: Geiler shice

codeofsumit (approved):

Falke-Design:
Changed the code.

Now it returns a FeatureGroup if `true` is passed.

```
map.pm.getGeomanLayers(true).toGeoJSON()
```

PR #686, opened 2020-10-20
Add option snappingOrder

### LINKED ISSUE #453: Snap on Point before Polyline
Hey, 

I am working with leaflet pm where I have multiple lines and points plotted on the map. The User can draw lines and is supposed to snap the line on the existing markers. The arrangement of my lines and points is such that there is always a point at the end of the line. Now, I want to register the snapping event which is working fine when snapping to a line, but when I snap to the end of the line where a point is placed it want to recognise the Point as the .layerInteractedWith compontent. However snapping always recognises the line instead. Is there a way to change this to a Point element?

Falke-Design:
Fix #453

### DIFF
--- a/src/js/Mixins/Snapping.js
+++ b/src/js/Mixins/Snapping.js
@@ -220,6 +220,7 @@ const SnapMixin = {
   },
   _calcClosestLayer(latlng, layers) {
     // the closest polygon to our dragged marker latlng
+    let closestLayers = [];
     let closestLayer = {};
 
     // loop through the layers
@@ -237,16 +238,22 @@ const SnapMixin = {
       // save the info if it doesn't exist or if the distance is smaller than the previous one
       if (
         closestLayer.distance === undefined ||
-        results.distance < closestLayer.distance
+        results.distance <= closestLayer.distance
       ) {
+        if(results.distance < closestLayer.distance){
+          closestLayers = [];
+        }
         closestLayer = results;
         closestLayer.layer = layer;
+        closestLayers.push(closestLayer);
       }
     });
 
+    closestLayers.find((result)=>result.layer instanceof L.Marker);
+
     // return the closest layer and it's data
     // if there is no closest layer, return undefined
-    return closestLayer;
+    return this._getOrderedSnappingLayer(closestLayers);
   },
   _calcLayerDistances(latlng, layer) {
     const map = this._map;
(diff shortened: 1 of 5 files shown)


src/js/Mixins/Snapping.js
```diff
@@ -237,16 +238,22 @@ const SnapMixin = {
...
       }
     });
 
+    closestLayers.find((result)=>result.layer instanceof L.Marker);
+
     // return the closest layer and it's data
-    // if there is no closest layer, return undefined
-    return closestLayer;
+    // if there is no closest layer, return an empty object
+    return this._getOrderedSnappingLayer(closestLayers);
```
codeofsumit, line 256: getClosestLayerByPriority

codeofsumit (changes requested):
function prioritise (key, order = 'asc') {
     const sortingOrder = {
       frankfurt: 1,
       xetra: 2,
       stuttgart: 3
     }

      return function (a, b) {
       if (!a.hasOwnProperty(key) || !b.hasOwnProperty(key)) return 0

        const first =
         a[key].toLowerCase() in sortingOrder
           ? sortingOrder[a[key]]
           : Number.MAX_SAFE_INTEGER
       const second =
         b[key].toLowerCase() in sortingOrder
           ? sortingOrder[b[key]]
           : Number.MAX_SAFE_INTEGER

        let result = 0
       if (first < second) result = -1
       else if (first > second) result = 1
       return order === 'desc' ? ~result : result
     }
   }

    // sort exchanges by priority
   exchanges.sort(prioritise('_id'))

src/js/Mixins/Snapping.js
```diff
@@ -338,6 +343,26 @@ const SnapMixin = {
...
+  _getClosestLayerByPriority(layers){
+    // sort the layers by creation, so it is snapping to the oldest layer from the same shape
+    layers = layers.sort((a,b)=>a._leaflet_id-b._leaflet_id);
+
+    const shapes = ['Marker','CircleMarker','Circle','Line','Polygon','Rectangle'];
+    const order = this._map.pm.globalOptions.snappingOrder || [];
+
+    let lastIndex = 0;
+    const prioOrder = {};
+    order.concat(shapes).forEach((shape)=>{
```
codeofsumit, line 355: add comment: 

// Merge user-preferred priority with default priority

codeofsumit (approved):

Issue #543, opened 2020-02-17
Adding marker in Circle with popups


cheazo:
Hey there, i'll try to add an marker onto an circle-layer which has an popup in it. But i can't.

Allways when i'll do a 'click' onto the circle to place the marker, then the popup comes up.
This happens in all differend shapes when i'll try to set a marker on them. But not when i'll try to draw an circle in an circle or in another shape. Only with markers.

Could you please check this out or tell me how to handle that?

Thx

codeofsumit:
@cheazo I'm unaware of your popup code, please create a JSfiddle showing the problem and I'm happy to help you out

Falke-Design:
Leaflet opens a popup if the clicked layer have one. [Leaflet src](https://github.com/Leaflet/Leaflet/blob/41706c01685320e3fd1d93d19e59b3275d3a90cb/src/layer/Popup.js#L390)

![markerpopup](https://s6.gifyu.com/images/marker_popup.gif)

It is working for other layer types, because they are having the hint markers and while clicking, the hint marker click event is triggered and not the event of the layer below. But if the hint marker is snapped, it is also not working.

Current state: https://jsfiddle.net/falkedesign/yu0925cf/
Fix one: Overwrite _openPopup from leaflet: https://jsfiddle.net/falkedesign/2opa4rwy/
Fix two: fixes only the marker problem, change iconAnchor: https://jsfiddle.net/falkedesign/Lypkgqo5/

Also I thinking of, if it is possible to have always a vanished marker or layer below the cursor and catching the click event.

codeofsumit:
@Falke-Design I think another invisible marker always attached to the mouse would be the cleanest solution. That would probably also be the base to fix multiple other problems 🤔

Falke-Design:
@codeofsumit a easier fix would be to use the event `mousedown`, it is triggered on the map, also when a popup is binded to a layer. What problems can happen if we change from `click` to `mousedown`?

PR #694, opened 2020-10-28
Disable Popup opening while drawing

Falke-Design:
The popup is unbinded while drawing and binded again after drawing is finished

Fix: #543

### DIFF
--- a/cypress/integration/rectangle.spec.js
+++ b/cypress/integration/rectangle.spec.js
@@ -158,5 +158,36 @@ describe('Draw Rectangle', () => {
       const coords = geojson.geometry.coordinates;
       expect(coords.length).to.equal(1);
     })
-  })
+  });
+
+  it('disable popup on layer while drawing', ()=>{
+    let rect = null;
+    cy.window().then(({ map, L }) => {
+      map.on("pm:create",(e)=>{
+        e.layer.bindPopup('Popup test');
+        if(e.layer instanceof L.Rectangle){
+          rect = e.layer;
+        }
+      });
+    });
+
+    cy.toolbarButton('rectangle').click();
+    cy.get(mapSelector)
+      .click(100,50)
+      .click(700,400);
+
+    cy.toolbarButton('marker').click();
+    cy.get(mapSelector)
+      .click(300,250);
+
+    cy.toolbarButton('edit').click();
+
+    cy.window().then(({ map}) => {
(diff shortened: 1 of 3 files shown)


cypress/integration/rectangle.spec.js
```diff
@@ -158,5 +158,36 @@ describe('Draw Rectangle', () => {
...
+          rect = e.layer;
+        }
+      });
+    });
+
+    cy.toolbarButton('rectangle').click();
+    cy.get(mapSelector)
+      .click(100,50)
+      .click(700,400);
+
```
codeofsumit, line 178: is this test really testing if the popup is disabled during through? I would expect a click on the layer here and then check that the popup is NOT there

src/js/Draw/L.PM.Draw.js
```diff
@@ -94,6 +95,17 @@ const Draw = L.Class.extend({
         map: this._map,
       });
     }
+
+    const layers = Utils.findLayers(this._map);
+    if(this._enabled){
+      layers.forEach((layer)=>{
+        Utils.disablePopup(layer);
+      })
+    }else{
```
codeofsumit, line 104: you need a formatter 😅

src/js/L.PM.Utils.js
```diff
@@ -42,6 +42,18 @@ const Utils = {
     }
     return L.polygon(polygon, circle.options);
   },
+  disablePopup(layer){
+    if(layer.getPopup()){
+      layer._pmCopyPopup = layer.getPopup();
```
codeofsumit, line 47: please rename `_pmCopyPopup` to `_tempPopupCopy`

codeofsumit (changes requested):
Looks good, some minor questions and name changes here

cypress/integration/rectangle.spec.js
Falke-Design: When the existing rectangle has a popup and we want to place a marker on it, it opens the popup instead of adding the marker. So it tests if the marker is placed and existing

codeofsumit (approved):

edudins:
By the way - for me this works only if I don't have `pmIgnore` set on the layers. But the problem I have is that I require `pmIgnore` so that my `readOnly` layers are not being accidently deleted or edited. Any ideas?

PR #696, opened 2020-10-30
Enable layers they are added while GlobalEditMode is active

Falke-Design:
Fix: #561 

old PR from you: #562


codeofsumit (approved):
nice

Issue #351, opened 2018-10-31
Display sq area/distance while drawing polygons/polylines/circles


m0t0r:
Hey @codeofsumit,

Thanks a lot for this project and your time!

I have been looking around the documentation and repo and was unable to find support for showing sq area/distance when drawing polygons/polylines/circles. I guess it's a feature request. I hope it can be added to this awesome library.

Thanks a lot!

codeofsumit:
Will close until this has more upvotes. But definitely something I want to add within the next months! Thank you for the suggestion.

sbilbault:
Hey @codeofsumit,

I would have the same feature request about distance/angle distance while drawing polygons/polylines. 

Is there a way to pass a custom function to pm while drawing each segment of a polygon, which would enable to display these parameters ? 

Thanks a lot,

codeofsumit:
@sbilbault I think there are other plugins that offer that functionality. leaflet-measure-path for example.

kunKun-tx:
`turf.js` works really well for calculating area. 
```
turf.area(layer.toGeoJSON())
```

bw1984:
the method described by @kunKun-tx works perfectly for polygons with distinct vertices. does anyone know if a way to calculate the area of a drawn circle? my existing function throws no errors however the geometry object which the circle generates only contains one coordinate pair (the center point) so the area is always calculated as 0. Any suggestions?

kunKun-tx:
@bw1984 
it's probably easier to calc the circle area using its raduis `layer.getRadius()`?

ref:
https://leafletjs.com/reference-1.6.0.html#circlemarker-getradius

codeofsumit:
I all, this feature is now implemented in [Leaflet-Geoman Pro](https://geoman.io/leaflet-geoman-pro) which is the way we fund the development of this open-source library. Eventually, features of the Pro version will trickle down to the OSS library which is happening already with many features. If you use Leaflet-Geoman in a professional environment, please consider going for the Pro version to get more features and support this open source project.

![measurement_hover](https://user-images.githubusercontent.com/2399810/98363061-a1332480-202e-11eb-8670-e2349a1d3377.PNG)
![measurement_options](https://user-images.githubusercontent.com/2399810/98363054-9f696100-202e-11eb-90da-36856e4830df.PNG)
![measurement_draw](https://user-images.githubusercontent.com/2399810/98363057-a09a8e00-202e-11eb-9cb4-a386b8ac77d0.PNG)

PR #711, opened 2020-11-23
Measurements

DimitryD:
Add calculating distance, perimeter and area for polyline and polygon
In drawing and editing mode
#351


codeofsumit:
Hi @DimitryD, thanks a lot for your work on this. We already built a measurement feature for Leaflet-Geoman which is currently only available through [Leaflet-Geoman Pro](https://geoman.io/leaflet-geoman-pro) - which funds the development of this open source library
So we have to close this PR as it's already part of the codebase and hopefully trickles down to the OSS version in the future. You can however release your work as a sub-plugin if possible for others to use.

PR #716, opened 2020-12-03
Add Norwegian translations

henrikmaa:


Falke-Design:
Can you please also add the file in: https://github.com/geoman-io/leaflet-geoman/blob/master/src/assets/translations/index.js

And the language code in the Readme: https://github.com/geoman-io/leaflet-geoman/blob/master/README.md#customize

PR #722, opened 2020-12-08
Prevent object to be deleted when removing vertecies.

### LINKED ISSUE #714: preventMarkerRemoval bug? - Prevent removal of first and last vertex on polyline and polygon?
Using preventMarkerRemoval right click to remove vertex, you can accidentally delete a Polyline or polygon.

For example Polyline requires 2 vertexes. If you right click it is "deleted" (actually still an object but cant be visible on map anymore because of missing position).

In my opinion it should not be allowed to remove first and last vertex of polyline?
The issue is the same for polygon if you dont have a total of 3 vertexes left. Not sure how to fix polygon tho..

Do you guys agree that this should be prevented?

henrikmaa:
Fixes #714 

Did not do anything about contextmenu on touch. which is still a [problem](https://github.com/geoman-io/leaflet-geoman/issues/714#issuecomment-738639409).


README.md
```diff
@@ -543,6 +543,7 @@ The following options are available globally and apply when going into global ed
 | pinning                   | `false` | Pin shared vertices/markers together during edit ⭐. [Details](#pinning)                                                     |
 | allowSelfIntersection     | `true`  | Allow/Disallow self-intersections on polygons and polylines.                                                                |
 | preventMarkerRemoval      | `false` | Disable the removal of markers/vertexes via right click.                                                                    |
+| preventObjectDelete       | `false` | Enable to disallow preventMarkerRemoval to remove the object. Requires preventMarkerRemoval: false                          |
```
codeofsumit, line 546: Let's rename this

Option Name: `removeLayerBelowMinVertexCount`
Default: true
Description: 
```
If `true`, vertex removal that cause a layer to fall below their minimum required vertices will remove the entire layer. If `false`, these vertices can't be removed. Minimum vertices are 2 for Lines and 3 for Polygons.
```

This requires a small change to your code to check for `removeLayerBelowMinVertexCount` to be `true` instead.

codeofsumit (approved):

Falke-Design:
Please merge https://github.com/henrikmaa/leaflet-geoman/pull/1 then we can merge this

codeofsumit (approved):
Good work @henrikmaa 👏  - thanks a lot for your work on this and congrats to your first merged PR at Leaflet-Geoman 🎉

Issue #719, opened 2020-12-07
Issue with Drawing Markers in geoman 2.8.0 - All layers are enabled on pm:create


henrikmaa:
After updating to Geoman 2.8.0 I have an issue that if i create a marker all other layers goes into editing mode. I cant reproduce this in a fiddle. 

What im doing exactly: 
 ```
map.pm.enableDraw('Marker', {
      snappable: true,
      snapDistance: 20,
      finishOn: 'click' || 'mousedown',
      continueDrawing: false <-- using this new addition of 2.8.0 (does not make a difference to my issue if i have it to true)
    });
```
Then on `pm:create` All layers are enabled for some reason

I think this may be an issue of rendering by using React, but I tried to remove all `pm.enable()` from my code, so basically it is just enabling by itself for no reason. I suspect there is maybe an issue in #696? Altough I am not using globalEditMode at all

If I was smart enough to understand properly whats happening  I would definitely give a better explanation...
Maybe it is not an issue with geoman at all, buy maybe you guys know what addition of 2.8.0 which may break this?

If I use geoman 2.7.0 it works fine.

Edit: Also tested with CircleMarker, which also makes all layers go into editing mode. If I create a Polyline or Polygon it works normal.

henrikmaa:
The problem was indeed a project code issue.

If anyone has something similar in the future:

I added the layer to LayerGroup on pm:create - `features.addLayer(e.layer)` not sure why it enables all layers tho.

Falke-Design:
I reopen this issue because it is a "bug" of the layergroup handling.

You can fix this when you disable the edit mode for markers they are drawn: `map.pm.setGlobalOptions({markerEditable: false})`


The problem is that by adding a layer to the layergroup, geoman is looking if a layer in the group is enabled (marker after placing) and if so all other layers will be also enabled. 
https://github.com/geoman-io/leaflet-geoman/blob/ff58d546ae5807a331485c16b5ac442b6bbce527/src/js/Edit/L.PM.Edit.LayerGroup.js#L31-L33

geohuz:
I'm encountering this issue today, wondering how can I get the lastest fix or have to wait the next version?

henrikmaa:
@geohuz fork geoman and use this commit until geoman has a new release :)

Falke-Design:
New Release is out

PR #726, opened 2020-12-15
Fix bug: when marker was drawn, the layergroup was enabled

Falke-Design:
Fix: #719


src/js/Edit/L.PM.Edit.LayerGroup.js
```diff
@@ -138,4 +139,19 @@ Edit.LayerGroup = L.Class.extend({
   getOptions() {
     return this._options;
   },
+  _getMap(){
+    if(this._map){
+      return this._map
+    }
+
+    for(let i = 0; i < this._layers.length; i+=1){
```
codeofsumit, line 147: Please use `array.find()` instead of this loop. This whole function could be a one-liner I think.



```js
_getMap() {
  return this._map || this._layers.find(l => !!layer._map)?._map || null
}

```

codeofsumit (changes requested):

PR #737, opened 2020-12-23
Add Persian translation (RTL)

Falke-Design:
Thanks to @SadraParandeh 

Note: this is a RTL Language #618

There is a Bug in the Leaflet Core Leaflet/Leaflet#7201 with RTL and Tooltips.
To work this have to be added in the CSS:
```
.leaflet-tooltip-pane > * {
    direction: rtl;
}
.leaflet-tooltip-pane {
    direction: ltr;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ Change the language of user-facing copy in leaflet-geoman
 map.pm.setLang('de');
 ```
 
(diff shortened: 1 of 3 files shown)


cksadra:
Many tanx...

PR #739, opened 2020-12-23
Replace complete .off() with specific one

Falke-Design:


src/js/Edit/L.PM.Edit.Line.js
```diff
@@ -85,10 +85,6 @@ Edit.Line = Edit.extend({
     poly.pm._markerGroup.clearLayers();
     poly.pm._markerGroup.removeFrom(this._map);
 
-    // clean up draggable
-    poly.off('mousedown');
-    poly.off('mouseup');
-
```
Falke-Design, line 91: I hope there are no problems if this lines are removed. I think this is a relict because it is implemented 2016. There was the edit and drag mode the same

PR #782, opened 2021-02-01
Add some new snapping options

elliots:
- Add snapSegment option to disable snapping to segments
- Add finishOn=snap option to finish shape when vertex is snapped
- Add requireSnapToFinish option

I can split these up if you prefer.


codeofsumit (changes requested):
Amazing options. Very clever I think they would make a great addition.
Would you be able to add some cypress tests as well? Otherwise we would jump in and add them to your PR.

README.md
```diff
@@ -243,14 +243,16 @@ See the available options in the table below.
 | snappable             | `true`                                | enable snapping to other layers vertices for precision drawing. Can be disabled by holding the `ALT` key.                                             |  
 | snapDistance          | `20`                                  | the distance to another vertex when a snap should happen                                                                                              |  
 | snapMiddle            | `false`                               | allow snapping in the middle of two vertices (middleMarker)                                                                                           |  
+| snapSegment           | `true`                                | allow snapping between two vertices                                                                                                                   |  
+| requireSnapToFinish   | `false`                               | require the last point of a shape to be snapped                                                                                                       |
```
codeofsumit, line 247:
```suggestion
| requireSnapToFinish   | `false`                               | require the last point of a shape to be snapped (only works for drawing mode)                                                                                                      |  
```

PR #783, opened 2021-02-01
Update typescript declaration file

elliots:
More complete, but not totally complete.

I also added some of the documentation to the file so it appears on mouseover in editors


ryan-morris:
@Falke-Design lgtm

codeofsumit:
Awesome guys, thanks a lot!

Issue #744, opened 2021-01-05
Module 'leaflet' resolves to an untyped module which cannot be augmented ^v2.8.0


lukekroon:
After upgrading my application from Angular 10 to Angular 11 and updating @geoman-io/leaflet-geoman-free from 2.7.0 to 2.9.0 I get the following error when building it with webpack:

```
Error: node_modules/@geoman-io/leaflet-geoman-free/dist/leaflet-geoman.d.ts:3:16 - error TS2665: Invalid module name in augmentation. Module 'leaflet' resolves to an untyped module at 'C:/***/node_modules/leaflet/dist/leaflet-src.js', which cannot be augmented.

3 declare module 'leaflet' {
                 ~~~~~~~~~
node_modules/@geoman-io/leaflet-geoman-free/dist/leaflet-geoman.d.ts:79:24 - error TS2304: Cannot find name 'ControlPosition'.

79             position?: ControlPosition;
                          ~~~~~~~~~~~~~~~
node_modules/@geoman-io/leaflet-geoman-free/dist/leaflet-geoman.d.ts:136:37 - error TS2304: Cannot find name 'PathOptions'.

136             setPathOptions(options: PathOptions): void;
```
Versions:
```
"@angular/core": "^11.0.5",
"@angular/cli": "^11.0.5",
"leaflet": "^1.7.1",
"@geoman-io/leaflet-geoman-free": "^2.9.0",
```

If I revert back to `"@geoman-io/leaflet-geoman-free": "2.7.0"` everything works perfectly. 

Please advise if this is a bug or a problem with something else.

Falke-Design:
In the release 2.8.0 we added Typescript typings #678. I think this is the cause of the problems, but @codeofsumit and I are both no typescript developer, so we don't know what is correct and how to fix this. Maybe you can upgrade the typings and make a PR.

Else you can try to remove the file `leaflet-geoman.d.ts` and changes from the PR #678 and look if it is now working. 

It would be very nice to hear from you, when you resolved this

lukekroon:
After reading [this ](https://stackoverflow.com/questions/42693836/typescript-custom-declaration-files-for-untyped-npm-modules/42702089)post I moved the the first line in `leaflet-geoman.d.ts` below the module declaration like this: 

```
declare module 'leaflet' {
    import * as L from 'leaflet';

    export interface GeomanLayerOptions {
        pmIgnore?: boolean;
    }
***
```

[...]

PR #820, opened 2021-03-19
Fixing typing issues on consuming in typescript project

soletan:
Regarding #744, I adjusted leaflet-geoman.d.ts as provided to prevent those errors. I see three issues with this file:

1. Importing from leaflet seems to have issues. However, this isn't affecting a consuming project by means of #744.
2. Declaring a module named `leaflet` seems to be wrong for your package isn't leaflet, but some extension. I guess that's what is basically confusing typescript in consuming projects for those projects have a dependency named leaflet and your code intends to declare its module or at least to extend that module's declaration. IMHO it is safe to declare your own module and make its types depending on leaflet by importing from their and exporting your stuff depending on the former.
3. There are two occurrences referring to type `L.PathOptions`, but in one case it is just `PathOptions`. And this is one of the errors causing #744. Adopting this I found `ControlPosition` to be actually `L.ControlPosition`. This is preventing #744, at least.

Using this version I can build my Stencil-based project.

For testing it, I had to fix the npm scripts for they are addressing binaries directly rather than trusting in npm looking for the proper scripts. I don't know why you've been addressing scripts directly, but this way `npm i` failed on Win10. If you have issues with my version of package.json on your system, what about using `npx` explicitly?


ryan-morris:
@soletan how are you importing leaflet in other files? This breaks import statements such as `import { Map, PM, LatLng } from 'leaflet';`  since it is being registered as a different module than leaflet now (why it declared the leaflet module). 

Redeclaring the module falls under the typescript spec as [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). The problem is with how leaflet plugins work, they register on top of existing types so it isn't necessarily the case here that we have a standalone library that just uses leaflet.

With the version that is in the current release, do you have `@types/leaflet` installed as well? Without including that package is the only way I've been able to reproduce the issue. It seems the build struggles since it is trying to assume typings based on the leaflet package files. With the explicit types included the module augmentation works. 

I think the solution here if including `@types/leaflet` fixes the build error for you, is to include `@types/leaflet` in this projects `package.json`

soletan:
I used to import leaflet like

```typescript
import * as L from 'leaflet';
```

in consuming modules. This way I was able to access definitions and was good to go without ever feeling a need to import @types/leaflet. Nonetheless, thanks for the hint. I'll try to add this dependency and see how it is affecting the previous issues first ...

soletan:
After importing that, the whole project went crazy for even Leaflet doesn't seem to have a clear concept for how to describe coordinates in their functions or at least related type definitions are more restrictive than accepted by the actual implementation. This is going to take a while to review ... gosh, I really hate Typescript ...

soletan:
Okay, forget about it. Type declarations require me to type cast lots of invocations because type definitions are too limited to see required features. AFAICT, this @types/leaflet is broken or tries to force people into either preventing some invocations they have to use for leaflet to work as expected or to flood their code with useless `(x as any)` type casts. I think, I'm better off without. Currently, all I need is a geoman library that's installable without manual modifications for some build tool to pass.

Issue #817, opened 2021-03-17
[Feature] Remove vertex on ...


Falke-Design:
Allow to remove a vertex / marker also with click or ctrl + click, dblclick, ... instead of contextmenu

Falke-Design:
This feature is added with the Release 2.11.0

PR #831, opened 2021-04-03
Add a few options to prevent adding / removing / moving of vertices and add Keyboard functions

Falke-Design:
New Edit-Mode options:
```
addVertexOn: 'click',
removeVertexOn: 'contextmenu',
removeVertexValidation: undefined, 
addVertexValidation: undefined,
moveVertexValidation: undefined
```

And add Keyboard functions `map.pm.Keyboard`:
> We implemented a built-in keyboard listener to make one central place where keyboard events can be accessed (without adding self a listener).
```
  getLastKeyEvent(type = 'current'),
  isShiftKeyPressed(),
  isAltKeyPressed(),
  isCtrlKeyPressed(),
  isMetaKeyPressed(),
  getPressedKey(),
```

Now we can add a check to remove / add / move vertices for example with ctrl hold:
```
const check = (obj)=>{
  return map.pm.Keyboard.isCtrlKeyPressed();
};
map.pm.setGlobalOptions({removeVertexValidation: check, addVertexValidation: check, moveVertexValidation: check})
```

Feature: #817


src/js/Mixins/Keyboard.js
```diff
@@ -0,0 +1,43 @@
...
+  _initKeyListener(map) {
+    this.map = map;
+    L.DomEvent.on(document, 'keydown keyup', this._onKeyListener, this);
+  },
+  _onKeyListener(e) {
+    let focusOn = 'document';
+
+    // TODO: .contains only supported since IE9, problem? Should we define, which Versions we are Support?
+    // with focusOn the user can add a check if the key was pressed while the user interacts with the map
+    if (this.map.getContainer().contains(e.target)){
```
Falke-Design, line 12: We will not support IE9

src/js/Mixins/Keyboard.js
```diff
@@ -0,0 +1,43 @@
...
+    const data = {event: e, eventType: e.type, focusOn};
+    this._lastEvents[e.type] = data;
+    this._lastEvents.current = data;
+
+    L.PM.Utils._fireEvent(this.map, 'pm:keyevent',data);
+  },
+  getLastKeyEvent(type = 'current') {
+    return this._lastEvents[type];
+  },
+  // TODO: isXYZPressed or getXYZPressed?
```
Falke-Design, line 25: we use `is...`

Falke-Design (commented):
Use in snapping the KeyboardMixin

src/js/Edit/L.PM.Edit.js
```diff
@@ -35,7 +40,48 @@ const Edit = L.Class.extend({
...
+      return false;
+    }
+    return true;
+  },
+  _vertexValidationDragEnd(marker){
+    if(marker._vertexResetLatLng){
+      marker._vertexResetLatLng = null;
+      return false;
+    }
+    return true;
```
codeofsumit, line 82: these validations don't seem to validate anything 🤔 
Also: where are the users validation functions being called?

codeofsumit (changes requested):

leaflet-geoman.d.ts
```diff
@@ -703,6 +742,8 @@ declare module 'leaflet' {
             allowSelfIntersection?: boolean;
         }
 
+        type VertexValidationHandler = (e: { layer: L.Layer, marker: L.Marker, event: any }) => boolean;
```
Falke-Design, line 745: is this correct? It should be a function with the arguments `layer`, `marker`, `event` and returns a boolean value.

leaflet-geoman.d.ts
```diff
@@ -716,12 +757,39 @@ declare module 'leaflet' {
...
+            addVertexOn?:
+                | 'click'
+                | 'dblclick'
+                | 'mousedown'
+                | 'mouseover'
+                | 'mouseout'
+                | 'contextmenu';
+
+            /** A function for validation if a vertex (of a Line / Polygon) is allowed to add. It passes a object with `[layer, marker, event}`. For example to check if the layer has a certain property or if the `Ctrl` key is pressed. (default:undefined). */
+            addVertexValidation?: undefined | VertexValidationHandler;
```
Falke-Design, line 776: is this working? undefined or the function

Falke-Design (commented):
@ryan-morris can you please look over this TypeScript implementation

leaflet-geoman.d.ts
ryan-morris: These 2 scenarios are valid based on these changes:

```
const b: L.PM.DrawModeOptions = {
    addVertexValidation: undefined,
};

const c: L.PM.DrawModeOptions = {
    addVertexValidation: e => { return false; },
};
```

Falke-Design: Ok thanks

codeofsumit (changes requested):

Issue #818, opened 2021-03-17
[Feature] Make LayerGroups draggable


Falke-Design:
Like in https://github.com/Falke-Design/PMDrag

Falke-Design:
This feature is added with the Release 2.11.0

PR #835, opened 2021-04-05
[Feature] Add LayerGroup dragging

Falke-Design:
Make LayerGroups draggable.

New option: `syncLayersOnDrag: true || Array[]`

Feature Request: #818


Falke-Design (commented):
syncLayerGroup

How is it looking, if a layer is in two laergroups

Falke-Design (commented):
Options `syncOnDrag: Array | true` Array of layers or LayerGroups, if true, all Layergroups are dragged, they contains this layer. -> `pm_parentLayerGroup`

PR #836, opened 2021-04-06
Refactor events and create new Mixin-Class for events

Falke-Design:
After refactoring 90 direct calls minified to 36 on `_fireEvent`.

Now we can be sure, that every event pass the same arguments.


Falke-Design (commented):
One Mixin for all. Only one event exists `centerplaced` -> same payload for edit and draw. Events that have undefined arguments, need to be explained why.

Description for event.

Add `source` from where event is fired

Add a arguement `custom Payload` and merge it with the current one

Falke-Design:
TODO: Add new Events f.ex. Rotation - merge first all other PRs

codeofsumit (approved):

PR #868, opened 2021-04-17
Fix #865

### LINKED ISSUE #865: TypeScript getGeomanDrawLayers improvement
I think the return type of the `getGeomanDrawLayers` method should be improved.  
Currently, in `leaflet-geoman.d.ts`, we have the following:

```ts
/** Returns all Geoman layers on the map as array. Pass true to get a L.FeatureGroup. */
getGeomanLayers(featureGroup?: boolean): L.FeatureGroup | L.Layer[];
/** Returns all drawn Geoman layers on the map as array. Pass true to get a L.FeatureGroup. */
getGeomanDrawLayers(featureGroup?: boolean): L.FeatureGroup | L.Layer
```

The issue is that `getGeomanDrawLayers` doesn't return an array (missing `[]` after `L.Layer`), so you have to rely on the `getGeomanLayers` method to perform array-specific operations (not ideal in more complex scenarios).

hstevanoski:
Fixed the return type of the `getGeomanDrawLayers()` method.


Falke-Design:
Thank you.

@codeofsumit ready to merge

PR #878, opened 2021-04-26
Rotation fix

G4mingEdition:


G4mingEdition:
Hi,

I'm using the "rotationV2" branch and I encountered a bug.
If you try to rotate a polygon with a hole, hole's points aren't rotating.
image

To fix the problem, I have edited "../js/Mixins/Rotating.js".
**Then**
image
**Now**
image

Falke-Design:
Thank you for reporting this issue and fixing it. I improved it a little bit, so that it is now also working if a polygon is cutted into two pices. Also I added some tests to check this in future.

@codeofsumit ready to merge

codeofsumit (approved):

Issue #821, opened 2021-03-22
Imcomplete type definition


soletan:
The current state of [leaflet-geoman.d.ts](https://github.com/geoman-io/leaflet-geoman/blob/master/leaflet-geoman.d.ts#L127) is causing issues with code complying with your documentation, mainly with regard to PM.Map. These methods are missing there at least:

* [controlsVisible()](https://github.com/geoman-io/leaflet-geoman#leaflet-geoman-toolbar)
* [removeControls()](https://github.com/geoman-io/leaflet-geoman#leaflet-geoman-toolbar)
* [reInitLayer()](https://github.com/geoman-io/leaflet-geoman#init-leaflet-geoman)
* [setGlobalOptions()](https://github.com/geoman-io/leaflet-geoman#draw-mode)

tirli:
@Falke-Design [reInitLayer()](https://github.com/geoman-io/leaflet-geoman/blob/master/leaflet-geoman.d.ts#L195) is still incomplete. It should accept layer as an argument.

Falke-Design:
@tirli  yes you are right ... will be fixed

PR #747, opened 2021-01-05
Update typescript interface

ocram1095:
While using it in typescript i found some missing functions in interface.

PR #881, opened 2021-04-29
improve typescript definitions

ryan-morris:


leaflet-geoman.d.ts
```diff
@@ -1,340 +1,779 @@
...
-            enableGlobalDragMode(): void;
-            disableGlobalDragMode(): void;
-            globalDragModeEnabled(): boolean;
-            toggleGlobalDragMode(): void;
-            enableGlobalRemovalMode(): void;
-            disableGlobalRemovalMode(): void;
-            globalRemovalModeEnabled(): boolean;
-            toggleGlobalRemovalMode(): void;
+            /** get shape */
+            getShape(): SUPPORTED_SHAPE_TYPES;
```
ryan-morris, line 758: should this be SUPPORTED_SHAPE (the string name)?

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,779 @@
...
-        export type GEOMAN_LAYER_CUT_MODE_EVENTS = 'pm:cut' | 'pm:edit';
-
-        export interface GeomanHelpers {
-            getShapes(): string[];
-            getActiveShape(): string;
+            | 'CircleMarker'
+            | 'ImageOverlay';
+
+        /** supported shape types. 'ImageOverlay' is in Edit Mode only. */
+        type SUPPORTED_SHAPE_TYPES =
```
ryan-morris, line 259: "Are here the Leaflet types correct? L.Marker because we define them only as String this._shape = 'Line';"

This is mainly used for the `shape` property on the event parameters. Are you saying those aren't set to the actual leaflet shape and just contain the shape name as a string?

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,779 @@
...
+
+            toggleStatus: boolean;
+        }
+
+        interface CustomControlOptions {
+            /** Name of the control */
+            name: string;
+
+            /** block of the control. 'options' is ⭐ only. */
+            block?: 'draw' | 'edit' | 'custom' | 'options';
```
ryan-morris, line 432: "block: is it possible to use BlockPositions? Else please add options ⭐"

this needs to be set to the string correct? `BlockPosition` is an object with properties

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,785 @@
...
+            ): void;
+
+            /** Set globalOptions and apply them. */
+            setGlobalOptions(options: GlobalOptions): void;
+
+            /** Apply the current globalOptions to all existing layers. */
+            applyGlobalOptions(): void;
+
+            /** Returns the globalOptions. */
+            getGlobalOptions(): GlobalOptions;
```
ryan-morris, line 337: are these the only ones that should be moved from PMDrawMap?

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,811 @@
...
+        }
+
+        type ACTION_NAMES = 'cancel' | 'removeLastVertex' | 'finish' | 'finishMode';
+
+        class Action {
+            text: string;
+            onClick?: (e: any) => void;
+        }
+
+        type TOOLBAR_CONTROL_ORDER =
```
ryan-morris, line 380: "I'm not sure if this line is correct: `changeControlOrder(order: 'drawCircle' | 'drawRectangle' | 'removalMode' | 'editMode'[]): void;` It can be each one of the "modes" :` drawMarker, drawCircleMarker, drawPolyline, drawRectangle, drawPolygon, drawCircle, editMode, dragMode, cutPolygon, removalMode` and every mode that is created with `createCustomControl` or `copyDrawControl` -> `CustomControlOptions.name`"

Like this?

ryan-morris:
This encompasses #865, merging that will likely result in unnecessary conflicts.

I think this also resolves #821 and #747

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,811 @@
...
+            toggle?: boolean;
+
+            /** Control is disabled. */
+            disabled?: boolean;
         }
 
-        export class MapDrawOptions {
+        interface GlobalOptions extends DrawModeOptions, EditModeOptions {
+            /** add the created layers to a layergroup instead to the map. */
+            layerGroup?: L.Map | L.LayerGroup;
```
Falke-Design, line 484: `snappingOrder`, `layerGroup` and `panes` are only on GlobalOptions available
https://github.com/geoman-io/leaflet-geoman#options

leaflet-geoman.d.ts
Falke-Design: Yes

Falke-Design (changes requested):
Complete Part "Rotation" is missing
- Draw option `rectangleAngle`
- Rotation: https://github.com/geoman-io/leaflet-geoman/tree/develop#rotate-mode


We will merge the typescript PRs from the other users so that they are listed as contributers, but we will overwrite it completly with your PR / file

ryan-morris:
@Falke-Design I think I covered all the areas you mentioned

ryan-morris:
> What is your plan how to deal with the events? It's not very clean, but for me that's enough for now.

Good question. I have a little more research I need to do. Primary issue at this moment is this:

```
    /**
     * Extends built in leaflet map.
     */
    interface Map {
        pm: PM.PMMap;

        on(type: 'pm:globaldrawmodetoggled', fn: (event: { enabled: boolean, shape: PM.SUPPORTED_SHAPES, map: L.Map }) => void, context?: any): L.Evented;
    }
```

putting the events on just the Map or just the Layer result in this TS2430 error:

```
Interface 'Map' incorrectly extends interface 'Evented'.
  Types of property 'on' are incompatible.
    Type '(type: "pm:globaldrawmodetoggled", fn: (event: { enabled: boolean; shape: string; map: Map; }) => void, context?: any) => Evented' is not assignable to type '{ (type: string, fn: LeafletEventHandlerFn, context?: any): this; (type: "baselayerchange" | "overlayadd" | "overlayremove", fn: LayersControlEventHandlerFn, context?: any): this; (type: "layeradd" | "layerremove", fn: LayerEventHandlerFn, context?: any): this; (type: "zoomlevelschange" | ... 18 more ... | "predrag"...'.
      Types of parameters 'fn' and 'fn' are incompatible.
        Types of parameters 'event' and 'event' are incompatible.
          Type '{ enabled: boolean; shape: string; map: Map; }' is missing the following properties from type 'LeafletEvent': type, target, sourceTarget, propagatedFrom, layerts(2430)
```

The Evented class in `@types/leaflet` immediately has a conflict. I do not think this should be happening. I've tried modifying the `Evented`  definition in `@types/leaflet` with no luck so that we could build on top of it. Adding directly to `Evented` class is the only way I can avoid errors, but `Evented` methods show up on both `Layer` and `Map` for Leaflet.

I have a couple more ideas but my Plan B is to reach out to the typescript gitter and stackoverflow, unless someone here has any suggestions.

Falke-Design:
@ryan-morris what if you use [LayerEvent](https://leafletjs.com/reference-1.7.1.html#layerevent) instead of Evented? Event Doku: [Events](https://leafletjs.com/reference-1.7.1.html#event-objects)

But it also looks like that map and layers are using `LayerEvent` together.

ryan-morris:
@Falke-Design `LayerEvent` describe the event being passed into the callback function, so that would result in:

```
layer.on('click', function(ev) {
    ev.on('pm:snap', ......);
});
```

which isn't a thing. The `.off()` do need to be added. Those didn't even cross my mind, so I'll get that updated. I haven't had a chance to look into the events in general any more, I am hoping to get a chance next week. Do you know roughly when the next release may be going out?

Falke-Design:
@ryan-morris This weekend I have a meeting with @codeofsumit to merge the PRs and prepare the release. But I think that some PRs needs a change. It would be good when we release around 14. of May but we will wait on you. This is a very important PR and improvement to the library.
So latest date for release is 21. of May

Falke-Design:
@ryan-morris can you please also add [.once()](https://leafletjs.com/reference-1.7.1.html#evented-once)? I hope this is the last one. I will make the review later

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+            actions?: (ACTION_NAMES | Action)[];
+
+            /** Control can be toggled. */
+            toggle?: boolean;
+
+            /** Control is disabled. */
+            disabled?: boolean;
+        }
+
+        type PANE = 'mapPane' | 'tilePane' | 'overlayPane' | 'shadowPane' | 'markerPane' | 'tooltipPane' | 'popupPane';
```
Falke-Design, line 560: PANE can also be a custom string. Please add ` | string`

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+            /** actions */
+            actions: (ACTION_NAMES | Action)[];
+
+            /** Function fired after clicking the control. */
+            afterClick: () => void;
+
+            /** CSS class with the Icon. */
+            className: string;
+
+            /**  */
```
Falke-Design, line 511:
```suggestion
            /** if true, other buttons will be disabled on click (default: true) */
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+            disableOtherButtons: boolean;
+
+            doToggle: boolean;
```
Falke-Design, line 514:
```suggestion
            /** Control can be toggled. */
            doToggle: boolean;
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+
+            jsClass: string;
```
Falke-Design, line 516:
```suggestion
            /** Extending Class f. ex. Line, Polygon, ... L.PM.Draw.EXTENDINGCLASS */
            jsClass: string;
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+
+            /** Function fired when clicking the control. */
+            onClick: () => void;
+
+            position: L.ControlPosition;
+
+            title: string;
```
Falke-Design, line 523:
```suggestion
            /** Text showing when you hover the control. */
            title: string;
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+
+            toggleStatus: boolean;
```
Falke-Design, line 525:
```suggestion
            /** toggle state true -> enabled, false -> disabled (default: false) */
            toggleStatus: boolean;
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
+
+            /** tool. 'options' is ⭐ only. */
```
Falke-Design, line 527:
```suggestion
            /** block of the control. 'options' is ⭐ only. */
```

leaflet-geoman.d.ts
Falke-Design, line 560:
```suggestion
        type PANE = 'mapPane' | 'tilePane' | 'overlayPane' | 'shadowPane' | 'markerPane' | 'tooltipPane' | 'popupPane' | string;
```

leaflet-geoman.d.ts
```diff
@@ -1,340 +1,1039 @@
...
-            afterClick?: (e: any) => void;
-            /** Action that appears as tooltip. Look under Actions for more information */
-            actions?: (Action | string)[];
-            /** Control can be toggled (default:true) */
-            toggle?: boolean;
-            /** Control is disabled (default:false) */
-            disabled?: boolean;
+        /** the position of each block. */
+        interface BlockPositions {
+            /** draw control position (default:''). */
```
Falke-Design, line 879:
```suggestion
            /** draw control position (default:''). '' also refers to this position */
```

Falke-Design (changes requested):
@ryan-morris it looks very well, i have found a few things but they are no big deal. If this is fixed, it is ready to merge.
Thank you a lot!

Falke-Design:
@ryan-morris do you think it is fixing this issues: #744, #821, #865?
Also I suggest that you create a new issue where you describe the problem with layer and map events and then I can add the label `help wanted`

ryan-morris:
@Falke-Design 

#744 I believe the answer to this is to have `@types/leaflet` installed. I do not see where someone would be using these typings but not the leaflet ones.

#821 + #865 - Yes.

Falke-Design (approved):

PR #829, opened 2021-03-31
[Feature] Make only certain layers editable / draggable / removeable / cutable

Falke-Design:
New options:
- allowEditing
- allowDragging
- allowRemoval
- allowCutting

Also there is a new option `layersToCut: [layer1, layer2, ...]` for the Cutting Mode, to cut only the layers in the array.

Fixes: #583, #276, #250, #420

PR #895, opened 2021-05-12
Add new functions / options to types

Falke-Design:
From PRs: #829 and #835

@ryan-morris can you please take a look into this?

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -481,7 +481,7 @@ declare module 'leaflet' {
             getControlOrder(): TOOLBAR_CONTROL_ORDER[];
 
             /** The position of a block (draw, edit, custom, options⭐) in the Toolbar can be changed. If not set, the value from position of the Toolbar is taken. */
-            setBlockPosition(block: 'draw' | 'edit' | 'custom' | 'options', position: ControlPosition): void;
+            setBlockPosition(block: 'draw' | 'edit' | 'custom' | 'options', position: L.ControlPosition): void;
 
             /** Returns a Object with the positions for all blocks */
             getBlockPositions(): BlockPositions;
@@ -723,7 +723,7 @@ declare module 'leaflet' {
             removeLayerBelowMinVertexCount?: boolean;
 
             /** Defines which layers should dragged with this layer together. true syncs all layers in the same LayerGroup(s) or you pass an `Array` of layers to sync. (default:false). */
-            syncLayersOnDrag?: [] | boolean;
+            syncLayersOnDrag?: L.Layer[] | boolean;
 
             /** Edit-Mode for the layer can disabled (`pm.enable()`). (default:true). */
             allowEditing?: boolean;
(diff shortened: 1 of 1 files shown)


leaflet-geoman.d.ts
```diff
@@ -722,6 +722,24 @@ declare module 'leaflet' {
             /** If true, vertex removal that cause a layer to fall below their minimum required vertices will remove the entire layer. If false, these vertices can't be removed. Minimum vertices are 2 for Lines and 3 for Polygons (default:true). */
             removeLayerBelowMinVertexCount?: boolean;
 
+            /** Defines which layers should dragged with this layer together. true syncs all layers in the same LayerGroup(s) or you pass an `Array` of layers to sync. (default:false). */
+            syncLayersOnDrag?: [] | boolean;
```
Falke-Design, line 726: @ryan-morris is this correct? It can be a boolean or an array of Layers,LayerGroups, FeatureGroups,GeoJSON Group,...

Falke-Design:
@ryan-morris should this not be `L.ControlPosition`? 

https://github.com/geoman-io/leaflet-geoman/blob/b1aeae9d3d1d5e01111e25345d84c37145cfcb6c/leaflet-geoman.d.ts#L882-L895

leaflet-geoman.d.ts
ryan-morris: If you know  all the allowed types to be in the array you can declare it like: 

`syncLayersOnDrag?: (L.Layer|L.FeatureGroup|L.GeoJSON)[] | boolean;`

if it's very open for what it will accept then you want:

`syncLayersOnDrag?: any[] | boolean;`

just specifying `[]` results in 

```
const a: L.PM.EditModeOptions = {
    syncLayersOnDrag: [layer,layer,featureGroup],
};

// error: Type 'Layer' is not assignable to type 'undefined'.ts(2322)
```

ryan-morris:
> @ryan-morris should this not be `L.ControlPosition`?
> 

Yes, those should be `L.ControlPosition`, do you want me to fix that in separate PR?

leaflet-geoman.d.ts
```diff
@@ -815,6 +833,9 @@ declare module 'leaflet' {
 
             /** angel of rectangle. */
             rectangleAngle?: number;
+
+            /** Cut-Mode: Only the passed layers can be cut. Cutted layers are removed from the Array until no layers are left anymore and cutting is working on all layers again. (Default: []) */
+            layersToCut?: [];
```
ryan-morris, line 838: This will have the same problem as `syncLayersOnDrag` like I mentioned. Should either be like `L.Layer[]` or `(L.Layer|L.SomethingElse)[]` or just `any[]`. Any implies passing things like numbers, or strings is ok, which I'm guessing definitely is not.

leaflet-geoman.d.ts
Falke-Design: Ok I understand. 
L.FeatureGroup, L.LayerGroup are extending from L.Layer, so does it works if I change it to `L.Layer[] | boolean`?

leaflet-geoman.d.ts
Falke-Design: Ok changed

leaflet-geoman.d.ts
ryan-morris: Yes

ryan-morris:
👍 these changes look good to me

Falke-Design:
Thank you for your timeand help

PR #902, opened 2021-05-16
Update Readme

Falke-Design:
Add missing functions and updated formatting

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -26,6 +26,12 @@ declare module 'leaflet' {
     interface Layer {
         pm: PM.PMLayer;
     }
+    /**
+     * Extends built in leaflet layergroup.
+     */
+    interface LayerGroup {
+        pm: PM.PMLayerGroup;
+    }
 
     interface Polygon {
         /** Returns true if Line or Polygon has a self intersection. */
@@ -696,7 +702,13 @@ declare module 'leaflet' {
             getActiveShape(): SUPPORTED_SHAPES;
 
             /** Set path options */
-            setPathOptions(options: L.PathOptions);
+            setPathOptions(options: L.PathOptions): void;
+
+            /** Set options */
+            setOptions(options: DrawModeOptions): void;
+
+            /** Get options */
+            getOptions(): DrawModeOptions;
         }
 
         interface CutModeOptions {
@@ -922,6 +934,9 @@ declare module 'leaflet' {
             /** sets layer options */
             setOptions(options?: EditModeOptions): void;
 
+            /** gets layer options */
+            getOptions(): EditModeOptions;
+
             /** Disables edit mode. */
             disable(): void;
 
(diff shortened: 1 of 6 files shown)


leaflet-geoman.d.ts
```diff
@@ -954,16 +969,42 @@ declare module 'leaflet' {
...
 
         interface PMLayer extends PMRotateLayer, PMEditLayer, PMDragLayer {
-            /** get shape */
+            /** get shape of the layer. */
             getShape(): SUPPORTED_SHAPES;
         }
 
+        interface PMLayerGroup {
+            /** Enables edit mode for all child layers. The passed options are preserved, even when the mode is enabled via the Toolbar */
+            enable(options?: EditModeOptions): void;
```
Falke-Design, line 978: @ryan-morris can you please check if this works, because the `enable` function of the LayerGroup has a private argument `_layerIds`. The user should only use the `options` argument
https://github.com/geoman-io/leaflet-geoman/blob/06c41aa1e8db8ea89f251eb25e4ad554d363250d/src/js/Edit/L.PM.Edit.LayerGroup.js#L54

leaflet-geoman.d.ts
```diff
@@ -26,6 +26,12 @@ declare module 'leaflet' {
     interface Layer {
         pm: PM.PMLayer;
     }
+    /**
+     * Extends built in leaflet layergroup.
+     */
+    interface LayerGroup {
+        pm: PM.PMLayerGroup;
+    }
```
Falke-Design, line 34: @ryan-morris does this works? I add the PMLayerGroup interface to the LayerGroup instead of the PMLayer interface

Falke-Design (commented):
@ryan-morris Thanks a lot for your help 🙏 We added you in the Readme.

leaflet-geoman.d.ts
ryan-morris: This actually is creating an error at that line,

```
Interface 'LayerGroup<P>' incorrectly extends interface 'Layer'.
  Types of property 'pm' are incompatible.
    Type 'PMLayerGroup' is missing the following properties from type 'PMLayer': getShape, enableRotate, disableRotate, rotateEnabled, and 8 more.ts(2430)
```

It shouldn't be needed. `LayerGroup` in `@types/leaflet` inherits from `Layer` so `pm` should have already been available, like:

```
this.pmLayers = new LayerGroup([]);
this.pmLayers.pm.disable();
```

leaflet-geoman.d.ts
ryan-morris: the typings for this just don't expose _layerIds as a parameter at all. So all the user should see as available is the options argument

leaflet-geoman.d.ts
ryan-morris: I misread your point and realize why you did that. Sorry I didn't respond sooner on this. That _should_ be on LayerGroup but it is conflicting with the `Layer` declaration of `pm`. I'm trying to see if there's a way around there. Error message goes away if I take `pm: PM.PMLayer;` off of `Layer` , but that is not a valid solution.

Falke-Design: Are you sure? because LayerGroups needs different functions on the pm object then default Layers

ryan-morris: Question:: We have `interface PMLayer extends PMRotateLayer, PMEditLayer, PMDragLayer` , should any of the methods listed on that or the *Layer it extends, be available on a `LayerGroup`?

Falke-Design: No, layergroup does not need / use any of the layer interfaces. So extending is not working

leaflet-geoman.d.ts
Falke-Design: Yes exactly

leaflet-geoman.d.ts
Falke-Design: https://www.reddit.com/r/typescript/comments/e0krdu/overriding_interface_type_declarations/

>You can't do that, however you can do this:
type IBar = Omit<IFoo, 'test'> & {
    test: Partial<IFoo['test']>;
};

Issue #912, opened 2021-05-18
Class 'LayerGroup<P>' incorrectly extends base class 'Layer'


luismiguelprs:
Hi, I'm getting this error on angular build process using leaflet + leaflet-geoman-free.

```
ERROR in node_modules/@types/leaflet/index.d.ts:1087:14 - error TS2415: Class 'LayerGroup<P>' incorrectly extends base class 'Layer'.
  Types of property 'pm' are incompatible.
    Type 'PMLayerGroup' is missing the following properties from type 'PMLayer': getShape, enableRotate, disableRotate, rotateEnabled, and 8 more.

1087 export class LayerGroup<P = any> extends Layer {
                  ~~~~~~~~~~
node_modules/@geoman-io/leaflet-geoman-free/dist/leaflet-geoman.d.ts:32:15 - error TS2430: Interface 'LayerGroup<P>' incorrectly extends interface 'Layer'.

32     interface LayerGroup {
                 ~~~~~~~~~~
```

Package versions: 

"@geoman-io/leaflet-geoman-free": "^2.11.1"
"leaflet": "^1.7.1"
"@types/leaflet": "^1.7.0"

Falke-Design:
Thanks for reporting this issue, we are on it.

@ryan-morris is there a way to fix this localy? A workaround until we release a new version.

ryan-morris:
I had to kept my project pinned at version `2.10.0` for the same reason. Outside of that, adding `"skipLibCheck": true` to tsconfig.json under `compilerOptions`, but I believe that will apply to all `*.d.ts` in node_modules.

I did not see a quick option solution to the problem either. In a similar way to [#893](https://github.com/geoman-io/leaflet-geoman/issues/893) the `@types/leaflet` inheritance complicates making 2 different types for same property name on `Layer` and `LayerGroup` since `LayerGroup` inherits from `Layer`. I'm still looking into options though

luismiguelprs:
Thanks @ryan-morris... I think the "skipLibCheck": true will resolve my problem for now..

Falke-Design:
#915 Should fix this issue

PR #915, opened 2021-05-19
Fix TypeScript Issues

Falke-Design:
The pm property is now added to L.Path instead of L.Layer. Now we can have different pm handlers for LayerGroup and Layers.

Also added some missing fields, more information in #917

Fixes: #912, #917

@ryan-morris please take a look if there is a reason, why this should not work.

leaflet_classes

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -21,12 +21,19 @@ declare module 'leaflet' {
     }
 
     /**
-     * Extends built in leaflet layer.
+     * Extends built in leaflet path.
      */
-    interface Layer {
-        pm: PM.PMLayer;
+    interface Path  {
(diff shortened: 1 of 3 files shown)


ryan-morris:
@Falke-Design so only problem is `pm` is not available on just a plain `Layer` reference now. This PR worked fine with my code, do you know how common a use case on `pm` for a simple layer is? [example](https://stackblitz.com/edit/typescript-abmw74?file=index.ts)

ryan-morris:
Ok, this is perfect solution then, and way it should have been originally.

contibru:
I`m facing the same issue here. 
I`m going to use geoman for the first time production and a got this error on ng build.
Glad it is fixed already.

codeofsumit (approved):

PR #925, opened 2021-05-29
Fix maxZoom === Infinity

Falke-Design:


codeofsumit (approved):
Interesting! didn't know there"s Infinity as a zoom level

Issue #950, opened 2021-07-06
With opt-in pmIgnored: false is required on map creation to set up map.pm. This is not specified in the docs.


main-kun:
https://github.com/geoman-io/leaflet-geoman/blob/dbf6c6cd5bdff79cf312fb324da8d979c38deb88/src/js/L.PM.js#L63-L66

When opt in is set to true before a map is created  `{ pmIgnored: false }` is required in options. Otherwise `map.pm` is undefined. This is not mentioned in the docs.

main-kun:
I've opened a pull request that adds a line about it in the docs: #951. However i was surprised to find out that when opt in is enabled newly drawn objects cannot be edited too. Is this by design? How an i enabled opt-in and at the same time allow editing new layers?

Falke-Design:
Yes this is the expected behaviour
https://github.com/geoman-io/leaflet-geoman#opt-in
> All layers will be ignored by Leaflet-Geoman, unless you specify pmIgnore: false on a layer

But you can add in your PR something like "layers will be also ignored if they are newly drawn"

PR #951, opened 2021-07-06
Add a line in the docs on pmIgnore with opt-in on map creation (#950)

main-kun:


README.md
```diff
@@ -142,6 +142,12 @@ And to disable it:
 L.PM.setOptIn(false);
 ```
 
+With opt-in you need to specify `pmIgnore: false` in map options on creation:
```
Falke-Design, line 145: Can you please change it to something like "If you have enabled opt-in before you init the map, you need to specify `pmIgnore: false` in the map options."

README.md
```diff
@@ -142,6 +142,12 @@ And to disable it:
...
+
+```js
+const map = L.map('map', { pmIgnore: true })
```
Falke-Design, line 148: Should it not be `pmIgnore: false`?

Falke-Design (changes requested):

README.md
main-kun: Ah yes pardon me

main-kun:
I amended the commit so that i don't pollute git history

README.md
```diff
@@ -142,12 +142,20 @@ And to disable it:
...
+const map = L.map('map', { pmIgnore: false })
+```
+
 All layers will be ignored by Leaflet-Geoman, unless you specify `pmIgnore: false` on a layer:
 
 ```js
 L.marker([51.50915, -0.096112], { pmIgnore: false }).addTo(map);
 ```
 
+Newly drawn layers will be ignored as well.
```
Falke-Design, line 157: Newly drawn layers will be ignored as well. 
To prevent this you can add this:

``` js
map.on('pm:create', (e) => {
  e.layer.setStyle({pmIgnore: false});
  L.PM.reInitLayer(e.layer);
});
```

Falke-Design (changes requested):
After adding my new change, this PR can be merged

README.md
main-kun: I took the liberty of changing the line to "To prevent this you can add the following event handler:" Double "this" sounds confusing.

Falke-Design (approved):
Ready to merge. Thank you

Issue #872, opened 2021-04-21
[Feature] Add Bezier curve function


Falke-Design:
Function to create a bezier curve

https://javascript.info/bezier-curve

#871

qpincon:
Hello,

Is there an appetite for this ? I made a [pull request](https://github.com/Leaflet/Leaflet.draw/pull/1008) into the leaflet-draw plugin, but it seems it is abandonned. 
The drawing and editing of bezier curves is made using the [Leafet.curve ](https://github.com/elfalem/Leaflet.curve) plugin. I made the integration into leaflet-draw so that the drawing of bezier curves can be enabled only if the Leaflet.Curve plugin is previously included.
I don't know if that's something you would be interesed in, and if if it's yes, in which way you would prefer to do it.
To check out what it looks like, you can download [this fork](https://github.com/qpincon/Leaflet.draw/tree/curve_built) and open `docs/examples/full.html`.

Cheers!

Falke-Design:
Yes of course this would be a nice feature. Because of the dependency of Leaflet.curve it would be better that it is supported as a add-on instead of adding it to the core library. Without Leaflet.curve we would add it to the core library.
The best would be to extend `L.PM.Draw.Line` and `L.PM.Edit.Line` and create a new shape.

Unfortunaly the demo on Leaflet-Draw Fork is not working. Can you create a video and add it here?

qpincon:
Sure, sorry about that, there seems to have some encoding issues with some of the files. Anyway, here are some gifs:
Drawing:
![draw](https://user-images.githubusercontent.com/10002904/131909983-990e0c7f-5229-40b4-a832-99751ca774fb.gif)

Editing:
![edit](https://user-images.githubusercontent.com/10002904/131910000-437649d6-c58f-469d-b8c4-e64ce1d6dcac.gif)

Falke-Design:
Looks nice. As I said if it is possible to add this feature without Leaflet.curve we will add it to Geoman. Maybe you can copy the nessery functions from Leaflet.curve. Else it would be a nice sub-plugin.

I will help you when ever you have question ;)

AndrejGajdos:
@qpincon is there any way you can provide a snippet what you have in gif?

PR #1026, opened 2021-11-05
Added curve drawing/editing support

qpincon:
As discussed in #872.

If the [Leaflet.curve](https://github.com/elfalem/Leaflet.curve) plugin is defined, the drawing of curve is enabled (with editing, dragging, rotation). 
Touch support is also handled, however not tested extensively.

Limitations:
At the moment there is not self-intersection detection for curves, and cutting the shape with the cut tool is not possible. It is a possible improvement in the future.

Let me know if you have any question.


Falke-Design:
Thanks for this nice feature! I have spoken with @codeofsumit and we decided that it makes sense to add this as a plugin, because it increases the package size over 100kb and we want that `Leaflet.curve` is added in the dependencies of package.json (so that it is working without that the user needs to include external Javascript)

If it is ok for you, we will setup a project `leaflet-geoman-curve` in the `geoman-io` organisation and then you can create a PR for the new repo.

qpincon:
Sure, no problem with that!

Falke-Design:
@qpincon new repo is created, can you please create a PR there: https://github.com/geoman-io/leaflet-geoman-curve

qpincon:
Thank you! 
I had to do some special case implementation for the handling of the rotation/dragging in the source code of those functions however. Should I Leave them here, or should I find a way to integrate them into the plugin you think?

Falke-Design:
Can we do it in following steps:
1. Make it possible to draw / edit
2. Make it possible to drag, rotate, ...

And I want to look where we need to implement API interfaces for plugins. So let us start with draw and edit ok?

AndrejGajdos:
@Falke-Design I tried the demo in development branch in https://github.com/geoman-io/leaflet-geoman-curve but it doesn't work.

Falke-Design:
@AndrejGajdos we currently stopped and never finished the development of leaflet-geoman-curve

PR #1028, opened 2021-11-07
Fix fire vertex removed event

lmachens:
In draw polyline mode, the `pm:vertexadded` event is fired like expected.
Clicking "Remove last vertex" doesn't fire a `pm:vertexremoved` event.

image

This pull request makes sure to fire this event.

### DIFF
--- a/src/js/Draw/L.PM.Draw.Line.js
+++ b/src/js/Draw/L.PM.Draw.Line.js
@@ -282,9 +282,16 @@ Draw.Line = Draw.extend({
       .filter((l) => !L.DomUtil.hasClass(l._icon, 'cursor-marker'))
       .find((l) => l.getLatLng() === removedCoord);
 
+
+    const markers = this._layerGroup.getLayers().filter((l)=>l instanceof L.Marker);
+    // the index path to the marker inside the multidimensional marker array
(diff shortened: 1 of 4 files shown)


Falke-Design:
A test would be good, do you think you can do this?

Falke-Design (approved):

lmachens:
Sorry, I was busy and just read your message. Your test looks good to me 👍

src/js/Draw/L.PM.Draw.Line.js
```diff
@@ -291,6 +291,8 @@ Draw.Line = Draw.extend({
     // sync the hintline again
     this._syncHintLine();
     this._setTooltipText();
+
+    this._fireVertexRemoved(marker, undefined, 'Draw');
```
codeofsumit, line 295:
```suggestion
    this._fireVertexRemoved(marker, undefined, 'Draw');
```

Let's check if we could/should send the old indexPath or an empty array instead of undefined

Falke-Design: To keep my first ideas
```    
const markers = this._layerGroup.getLayers();
    // the index path to the marker inside the multidimensional marker array
    const { indexPath } = .findDeepMarkerIndex(
      markers,
      marker
    );
```

PR #1046, opened 2021-12-05
Fix dragging in Canvas mode

### LINKED ISSUE #1023: ImageOverlay not being dragged in latest release
In version 2.8.0 you can drag ImageOverlays  just fine (https://jsfiddle.net/830xkbLv/)
But it does not seems to be working in 2.11.3 (https://jsfiddle.net/em806a14/13/)

### LINKED ISSUE #1027: Drag mode tool doesn't work with Canvas
After drawing some geometries on the map, any attempt to move them using the drag mode tool results in the map view moving instead of the intended geometry. This appears to be caused by the use of leaflet's canvas mode, because if you remove the `preferCanvas` property from the `MapContainer` component, the bug disappears.

https://stackblitz.com/edit/react-bbsept?file=src/App.js

Although I'm using `react-leaflet`, I basically assumed it's not a factor, and I realize I could be wrong about that. I haven't done the same test without react.

Falke-Design:
After fixing dragging for touch devices in the Release 2.11.0, dragging was not supported anymore for Canvas mode.

fix #1023, fix #1027

### DIFF
--- a/cypress/integration/imageoverlay.spec.js
+++ b/cypress/integration/imageoverlay.spec.js
@@ -33,4 +33,36 @@ describe('Opens Testing Environment', () => {
       expect(eventcalled).to.equal('pm:snap');
     });
   });
+
+  it('Drags ImageOverlay', () => {
+    let eventcalled = false;
+    let io;
+    cy.window().then(({ map, L }) => {
+      map.setView([18.74469, 72.1258], 10);
+      const icon =
+        'https://camo.githubusercontent.com/33fa9a94048274f81a806631ca881a55c2aa8f0a/68747470733a2f2f66696c652d6a787a796a67717775742e6e6f772e73682f';
+      io = L.imageOverlay(
+        icon,
+        [
+          [18.74469, 72.1258],
+          [18.94469, 72.3258],
+        ],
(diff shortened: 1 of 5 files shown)


pblairce:
> After fixing dragging for touch devices in the Release 2.11.0, dragging was not supported anymore for Canvas mode.
> 
> fix #1023, fix #1027

This change break geoman for me... this line here https://github.com/geoman-io/leaflet-geoman/blame/1fdc918fa39ffa84327fdf639fa75865168f716d/src/js/Mixins/Dragging.js#L59 causes errors on the desktop version

pblairce:
> > After fixing dragging for touch devices in the Release 2.11.0, dragging was not supported anymore for Canvas mode.
> > fix #1023, fix #1027
> 
> This change break geoman for me... this line here https://github.com/geoman-io/leaflet-geoman/blame/1fdc918fa39ffa84327fdf639fa75865168f716d/src/js/Mixins/Dragging.js#L59 causes errors on the desktop version

this._map is undefined on the desktop

Falke-Design:
Thx for reporting, there is already an issue for this #1069

PR #1056, opened 2021-12-18
WIP: Better support to create Plugins and new Shapes

Falke-Design:
There are a lot of problems if we want to create a plugin for Leaflet-Geoman. We can access the Shape filters, the rotation and snapping is not modular and it is not easy possible to insert new Translations for existing languages


mscno:
Closing this long-stale WIP to keep the PR queue actionable. It can be recreated if the plugin API work resumes.

PR #1058, opened 2021-12-18
Run formatter

Falke-Design:


src/js/helpers/index.js
```diff
@@ -189,7 +194,8 @@ export function prioritiseSort(key, _sortingOrder, order = 'asc') {
   let objKey;
   let n = keys.length;
   const sortingOrder = {};
-  while (n--) {
+  while (n > 0) {
+    n -= 1;
```
codeofsumit, line 198: This means that the while loop will run with the value 0

When n is 1, we jump in the keep. Then the value is reduced to 0 and the code runs. Is that what we want?

If the reason we do this at the beginning is the starting value, we should reduce the starting value in the declaration instead and keep the reduction in the loop at the end.

codeofsumit (changes requested):
There might bea problem with the while loop

src/js/helpers/index.js
Falke-Design: Few tests:
```
n = 5;
while(n--){
	console.log(n)
}
// 4,3,2,1,0
```
```
n = 5;
while(n > 0){
	n -= 1
	console.log(n)
}
// 4,3,2,1,0
```
```
n = 4;
while(n > 0){
	console.log(n)
	n -= 1
}
// 4,3,2,1,0
```

PR #1084, opened 2022-01-21
Remove Node Engine. Fixes #1082 (patch)

### LINKED ISSUE #1082: NodeJS engine version >= 16 specifier
Hi, first of all thanks for this module, which does a good job. I just want to highlight that this [recent commit](https://github.com/geoman-io/leaflet-geoman/commit/1a47ff13533aeefef90b734e7d2a485c0db5b182#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R21) has broken our build, although it has been released as a patch. Indeed, you have added a specifier on the node engine version that does not allow to install the module with versions prior to 16.

Of course I understand NodeJS v16 should be the target version because it is the LTS version, but releasing such a change as a patch can probably raises some weird issues.

codeofsumit:
Enforcing a node version broke user implementations instead of its intended use: enforcing node version 16 among contributors. So, we are reverting this for now. Fixes #1082

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@geoman-io/leaflet-geoman-free",
   "version": "2.11.4",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@geoman-io/leaflet-geoman-free",
-      "version": "2.11.2",
+      "version": "2.11.4",
       "license": "MIT",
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):

max-castro:
Hi @codeofsumit do you have any plans on releasing a new version with this? We found an issue with 2.11.3 that works on 2.11.4 but updating our CI with 16 is not feasible right now

PR #1090, opened 2022-01-28
Fixed redirect to another web page when clicking in the disabled control button

vvlladd28:
When the control buttons are disabled, need to stop the propagation event because the call default behavior hyperlink scroll top or redirects to another page in SPA.


vvlladd28:
Please review PR @Falke-Design

Falke-Design (approved):
Good catch!

PR #1094, opened 2022-02-02
Improvement disable control button when change state we do not redraw panel

vvlladd28:
When studying the code, it was found that the optimal movement, which causes a complete redrawing of the control panel, when changes disable state.

This PR makes the following changes: 

- When we change disable state, we change CSS class and internal state
- When you lock a button, we forcibly render it inactive
- Changed CSS styles to remove class duplication for different DOM element

### DIFF
--- a/cypress/integration/toolbar.spec.js
+++ b/cypress/integration/toolbar.spec.js
@@ -413,9 +413,6 @@ describe('Testing the Toolbar', () => {
   });
   it('Disable active button', () => {
     let eventFired = '';
-
-    cy.toolbarButton('polygon').click();
-
     cy.window().then(({ map }) => {
       map.on('pm:buttonclick', ({ btnName }) => {
         eventFired = btnName;
--- a/cypress/integration/toolbar.spec.js
+++ b/cypress/integration/toolbar.spec.js
@@ -408,7 +408,31 @@ describe('Testing the Toolbar', () => {
     cy.toolbarButton('polygon').click();
 
     cy.window().then(() => {
-      expect(eventFired).to.not.equal('drawPolygon');
+      expect(eventFired).to.equal('drawPolygon');
+    });
+  });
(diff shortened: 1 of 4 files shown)


vvlladd28:
Please review my PR @Falke-Design

cypress/integration/toolbar.spec.js
```diff
@@ -408,7 +408,28 @@ describe('Testing the Toolbar', () => {
...
+
+    cy.window().then(({ map }) => {
+      map.pm.Toolbar.setButtonDisabled('drawPolygon', true);
+    });
+
+    cy.window().then(() => {
+      expect(eventFired).to.equal('drawPolygon');
+      cy.toolbarButton('polygon')
+        .closest('.button-container')
+        .should('have.not.class', 'active');
```
Falke-Design, line 432: Move this block above
```
 cy.window().then(({ map }) => {
      map.pm.Toolbar.setButtonDisabled('drawPolygon', true);
    });
```
And then add a check if the event / click happens after disableing the button

Falke-Design (changes requested):
Why do you think it is important to merge this? Clearly the difference is that the Toolbar is not rendered new, but every change on the Toolbar cause a re-render so why should we handle this different?

Anyway thank you for your contribution.

vvlladd28:
For example, I can give an example of [code](https://github.com/Leaflet/Leaflet/blob/4f639a85efffa49c3e64a07dc0b6f5aa73f13449/src/control/Control.Zoom.js#L102) from the leaflet library, they have buttons for zooming the map and don't render after disable. The redraw render is complex for the browser.

cypress/integration/toolbar.spec.js
vvlladd28: Thank you, fixed.

codeofsumit (approved):

PR #1159, opened 2022-05-02
Drag layers mobile device

BaharWeb:
fix the issue of dragging layers while using Geoman on mobile devices


BaharWeb:
unfortunately this solution could not resolve the issue of dragging layer on mobile devices

Falke-Design:
@BaharWeb Thank you for investigating. I'm already working on a fix

BaharWeb:
@Falke-Design Thank you

PR #1176, opened 2022-06-14
Avoid empty clientX and clientY values when dragging Marker (Safari iOS)

wvddrss:
On mobile safari dragging a Marker gave an error. 
When digging into the code I saw that it calls:
`this._map.mouseEventToContainerPoint(e.originalEvent);`
But originalEvent did not contain a `clientX` and `clientY` property.
In the call `_simulateMouseDownEvent` I've used the later declared first touch event, which has a `ClientX|Y` value.


Falke-Design:
@wvddrss can you please revert the last commit. It looks like it doesn't belongs to here

wvddrss:
My apologies! Well noticed! Made a new PR with seperate branch.

PR #1208, opened 2022-07-14
Avoid empty clientX and clientY values when dragging marker in mobile…

wvddrss:
On mobile safari dragging a Marker gave an error. 
When digging into the code I saw that it calls:
`this._map.mouseEventToContainerPoint(e.originalEvent);`
But originalEvent did not contain a `clientX` and `clientY` property.
In the call `_simulateMouseDownEvent` I've used the later declared first touch event, which has a `ClientX|Y` value.


Falke-Design:
I think the other simulations should have the same workaround (`_simulateMouseMoveEvent`, `_simulateMouseUpEvent`)

codeofsumit (approved):

PR #1215, opened 2022-08-22
Type Return Improvements

TurtIeSocks:
- Adds overrides for both `getGeomanLayers` and `getGeomanDrawLayers`, giving better IDE feedback when writing TypeScript.
- Adds TS support for setting a custom radius when creating CircleMarkers

Screen Shot 2022-08-22 at 7 01 46 PM


Falke-Design (approved):
Thank you!

PR #1216, opened 2022-08-24
Option for setting/editing the circle radius

TurtIeSocks:
- New `radiusEdit` option
- Makes circles behave similarly to circleMarkers when drawing and editing when set to `false`
- Defaults to `true` to maintain backwards compatibility

Admittedly, this is a lot of copy/paste from the CircleMarker counterpart classes. From my testing, everything should behave as it did before without setting `radiusEdit`, but would appreciate a second pair of eyes.

Goal:
Be able to draw circles without having to fuss with the radius, while maintaining their size on the map when zooming, since CircleMarker's scale with zoom.


src/js/Mixins/Dragging.js
```diff
@@ -364,7 +364,7 @@ const DragMixin = {
       });
 
     if (
-      this._layer instanceof L.Circle ||
+      (this._layer instanceof L.Circle && this._layer.options.editable) ||
```
Falke-Design, line 367: Should it not be `radiusEdit`?

README.md
```diff
@@ -288,6 +288,7 @@ See the available options in the table below.
 | minRadiusCircleMarker | `null`                                | Set the min radius of a `CircleMarker` when editable is active.                                                                                                                                                 |
 | maxRadiusCircleMarker | `null`                                | Set the max radius of a `CircleMarker` when editable is active.                                                                                                                                                 |
 | editable              | `false`                               | Makes a `CircleMarker` editable like a `Circle`.                                                                                                                                                                |
+| radiusEdit            | `true`                                | Enables radius editing after drawing a `Circle`
```
Falke-Design, line 291: Why not using `editing` like for CircleMarker?

src/js/Draw/L.PM.Draw.Circle.js
```diff
@@ -252,13 +330,18 @@ Draw.Circle = Draw.extend({
...
     }
 
-    if (this.options.minRadiusCircle && radius < this.options.minRadiusCircle) {
-      radius = this.options.minRadiusCircle;
-    } else if (
-      this.options.maxRadiusCircle &&
-      radius > this.options.maxRadiusCircle
-    ) {
-      radius = this.options.maxRadiusCircle;
+    if (this.options.radiusEdit) {
```
Falke-Design, line 333: Is always true because `_finishShape()` is only called if `radiusEdit` is true

src/js/Draw/L.PM.Draw.js
```diff
@@ -65,7 +65,7 @@ const Draw = L.Class.extend({
     this.shapes.forEach((shape) => {
       this[shape] = new L.PM.Draw[shape](this._map);
     });
-
+    this.Circle.setOptions({ radiusEdit: true, continueDrawing: true });
```
Falke-Design, line 68:
```suggestion
    this.Circle.setOptions({ radiusEdit: true });
```

src/js/Edit/L.PM.Edit.Circle.js
```diff
@@ -10,6 +10,9 @@ Edit.Circle = Edit.extend({
     this._updateHiddenPolyCircle();
   },
   enable(options) {
+    // backwards compatibility
+    options.allowEditing = options.radiusEdit ?? true;
```
Falke-Design, line 14: `allowEditing` enables the editing in generally, so this line should be remove. Editing with `radiusEdit: false` should be the same as dragging but in editMode

Falke-Design (changes requested):
Thank you for your contribution. I will take a deeper look in the next 2 weeks. I have currently to much to do.

src/js/Mixins/Dragging.js
TurtIeSocks: Oops, leftover fragment from when this PR first started using `editable` as the option.

README.md
TurtIeSocks: PR started with using `editable` like CircleMarker does but I ran into an issue when trying to set the backwards compatibility value, with it then being set for both and leading to unwanted results. In the end, it seemed like it'd be best to separate them, in case anyone only wants one or the other.

src/js/Edit/L.PM.Edit.Circle.js
TurtIeSocks: I must have been staring at this PR for too long... should be (and has been fixed in latest commit) `options.radiusEdit = options.radiusEdit ?? true`

src/js/Draw/L.PM.Draw.Circle.js
TurtIeSocks: Fixed in latest commit.

TurtIeSocks:
Thanks for the review and comments! Fixed or explained all of the comments in latest commit.

src/js/Edit/L.PM.Edit.Circle.js
TurtIeSocks: If there's somewhere better to set this like there was with L.PM.Draw, let me know!

TurtIeSocks:
@Falke-Design sorry for the tag but anything I can do to move this along?

README.md
Falke-Design, line 291: I'm thinking of naming it `radiusEditCircle` then we can have in future `radiusEditCircleMarker` with a different value.

Falke-Design, line 291:
```suggestion
| radiusEdit            | `true`                                | Enables radius editing while drawing a `Circle`
```

leaflet-geoman.d.ts
```diff
@@ -1062,6 +1062,9 @@ declare module 'leaflet' {
       /** Makes a CircleMarker editable like a Circle (default:false). */
       editable?: boolean;
 
+      /** Edit a circle's radius after placement (default:true). */
```
Falke-Design, line 1065:
```suggestion
      /** Enables radius editing while drawing a Circle (default:true). */
```

Falke-Design (changes requested):
I need more time to review it. I hope in the next 2 weeks.

Sorry for the long waiting!

TurtIeSocks:
No worries, @Falke-Design, I totally get it :) just wanted to check in. I've resolved all of your latest comments.

TurtIeSocks:
bump

Falke-Design:
Superseded by #1216

PR #1230, opened 2022-09-23
optimize isEmptyDeep

drzhbe:
**TL;DR** `isEmptyDeep` is now 300 times faster

Function `isEmptyDeep` suffered from 2 problems:
1. Unnecessary memory allocations
2. No early return

And since it's used in the snapping code, whenever you have many or just big geometries, the geometry editing is janky.

In this PR I optimized the `isEmptyDeep` function and added the tests for it.

Video explanation:
https://www.loom.com/share/e70379b76e5141d984e773b70dd47921

Benchmark:
https://jsbench.me/h8l8dv6d7h/2

### DIFF
--- a/cypress/integration/helpers.spec.js
+++ b/cypress/integration/helpers.spec.js
@@ -1,35 +1,35 @@
-import { isEmptyDeep } from '../../src/js/helpers';
+import { hasValues } from '../../src/js/helpers';
 
-describe('Helper isEmptyDeep', () => {
-  it('should return true for empty arrays', () => {
-    expect(isEmptyDeep([])).to.be.true;
+describe('Helper hasValues', () => {
+  it('should return false for empty arrays', () => {
+    expect(hasValues([])).to.be.false;
   });
 
-  it('should return true for empty nested arrays', () => {
-    expect(isEmptyDeep([[[]]])).to.be.true;
+  it('should return false for empty nested arrays', () => {
+    expect(hasValues([[[]]])).to.be.false;
   });
 
-  it('should return true for empty-like arrays', () => {
-    expect(isEmptyDeep([undefined, null, ''])).to.be.true;
+  it('should return false for empty-like arrays', () => {
+    expect(hasValues([undefined, null, ''])).to.be.false;
   });
 
-  it('should return true for empty-like nested arrays', () => {
-    expect(isEmptyDeep([[[undefined], [null], ['']]])).to.be.true;
+  it('should return false for empty-like nested arrays', () => {
+    expect(hasValues([[[undefined], [null], ['']]])).to.be.false;
   });
 
-  it('should return false for non-empty arrays', () => {
-    expect(isEmptyDeep([0])).to.be.false;
+  it('should return true for non-empty arrays', () => {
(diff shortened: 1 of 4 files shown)


cypress/integration/helpers.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(isEmptyDeep([[[undefined], [null], '']])).to.be.true;
+  });
+
+  it('should return false for a non-empty nested arrays', () => {
+    expect(isEmptyDeep([[[1]]])).to.be.false;
+  });
+
+  it('should return false if there is at least one non-empty value', () => {
+    expect(isEmptyDeep([[[null], [undefined, 1]]])).to.be.false;
+  });
```
Falke-Design, line 22: Can you please add tests with no nested arrays `isEmptyDeep([null, undefined, 1])` and `isEmptyDeep([null, undefined, ''])`

src/js/helpers/index.js
```diff
@@ -13,13 +13,22 @@ export function getTranslation(path) {
...
+  const hasValues = (list) => {
+    for (let i = 0; i < list.length; i += 1) {
+      const item = list[i];
 
-  return !flatten(l).length;
+      if (Array.isArray(item)) {
+        if (hasValues(item)) {
+          return true;
+        }
+      } else if (item != null && item !== '') {
```
Falke-Design, line 24: Where is `undefined` checked? I see that the tests are working but I don't get where `undefined` is handled 😄

Falke-Design (changes requested):
Wow that is great! Thank you!

Please add a few more tests like requested and then I think we can merge this

src/js/helpers/index.js
drzhbe: `a == null` is the same as `a === undefined || a === null`. I can make it explicit

Falke-Design: ah yeah sure, I didn't noticed that it only has one `=`. Yes please make it explicit, thanks

drzhbe:
@Falke-Design I've added the requested tests and made `=== undefined && === null` explicit
The tests look pretty tidy to me, wdyt?
Screen Shot 2022-09-23 at 9 35 18 pm

Falke-Design (approved):
This is really a nice improvement! Thank you very much!

@codeofsumit will make merge before the next release

drzhbe:
@Falke-Design Thank you for the review! Do you know when approximately next release is gonna happen? Is it weeks/months away?

Falke-Design:
Maybe with the end of oktober. I need first to fix most of the bugs but currently I have not much time

drzhbe:
@Falke-Design I simplified the function a bit, removed the inner function `hasValues`. Now `isEmptyDeep` itself is recursive.

Falke-Design:
I'm thinking about, if we should rename the function to `hasValues`, it would make it easier to understand the code in the function. Or does it not matter? What do you @drzhbe think?

drzhbe:
@Falke-Design I agree the function reads a bit more straightforward now. And the name is shorter. Good suggestion 👍

codeofsumit (approved):
Amazing !!

PR #1260, opened 2022-11-28
Add setInitAngle

### LINKED ISSUE #1245: Cannot add a Rectangle with a rotation angle from a database.
My use case involves drawing a rectangle on the map and then having the ability to rotate them and save them as a GeoJson feature in order to be loaded into the map at a later time.

You are able to draw a rectangle layer of course using ( L.rectangle(*rectangle bounds*) ) but there is no way to add the rectangle with a set angle, it will just load with an angle of 0°. 

I can take those saved rectangles and create a FeatureCollection, and load those into a GeoJson Layer (L.geoJson(collection)) which works really well with rotating, dragging, etc., but doesn't work with editing. Because it's a GeoJson the Geoman editing tool just sees it as a Polygon and tries to edit it as such.

I do realize this is kind of a core leaflet issue with the way layers are drawn and identified.

This also seems to be somewhat related to another issue that can be seen on [https://geoman.io/geojson-editor](https://geoman.io/geojson-editor)

- draw a rectangle
- rotate the rectangle
- and then edit the rectangle

The rectangle will snap back to 0° because it's a rectangle layer and is just using the rectangle bounds to update/redraw that layer again.

Falke-Design:
Problem: 
After importing a rotate layer for example a rectangle the angle needs to be set initial.

angle-import

Fixes #1245


README.md
```diff
@@ -632,6 +632,7 @@ The following methods are available for layers under `layer.pm`:
 | rotateLayer(`degrees`)        | -         | Rotates the layer by `x` degrees.                |
 | rotateLayerToAngle(`degrees`) | -         | Rotates the layer to `x` degrees.                |
 | getAngle()                    | `Degrees` | Returns the angle of the layer in degrees.       |
+| setInitAngle(`degrees`)       | -         | Set the initial angle of the layer in degrees.   |
```
Falke-Design, line 635: Add in the Readme a note about Rectangles + Geoman-www fix

codeofsumit (approved):

PR #1262, opened 2022-11-30
Add Finnish translations

masysma:
Add Finnish translations, import them and add 'fi' as supported language in README.md


Falke-Design (changes requested):
Please add the language code in `leaflet-geoman.d.ts` too

masysma:
Requested changes done.

Falke-Design (approved):

PR #1271, opened 2022-12-17
bump deps to fix node 18 dev compatibility

TurtIeSocks:


codeofsumit:
Please use node 16.x for development. We've added it to package.json and .nvmrc now.
Updating the project and it's deps to v18 is something we're planning for next year.

Issue #1300, opened 2023-01-30
Map is not draggable after adding a text


lmachens:
Hey 👋,

I tried out the text layer functionally and really like it.
But after adding a text, my map is not draggable anymore.

The same issue is reproduceable on https://geoman.io/leaflet-geoman.

Am I missing something?

![draggable](https://user-images.githubusercontent.com/10058950/215443351-f4a9b685-88ae-4d96-8c2f-b578afa2c12b.gif)

As workaround, I set `map.dragging.enable()` after using the tool.

Falke-Design:
Thank you for reporting! Will be fixed in #1302

PR #1302, opened 2023-02-01
Fix map dragging after editing Text-Layer

Falke-Design:
Fixes: #1300


cubecleveland:
Im still seeing this issue in Geoman 2.14.1
Is that fixed ?

PR #1309, opened 2023-02-11
[Refactor] Extends Circle from CircleMarker (patch)

Falke-Design:
In general the code for `Circle` and `CircleMarker` was the same. To reduce the maintainence work, this PR makes that the `Circle` extends from `CircleMarker` and overwrites only the necessary options and functions.

Additionally it is now possible to hide the radius edit-handlers with `editableCircle: false` like for the CircleMarker. 
And the option of CircleMarker is renamed from `editable` to `editableCircleMarker`, with a fallback for older versions.

- [x] TODO: find a better name then `editable`

Supersede #1216


Falke-Design:
@TurtIeSocks sorry for the **very** late replay! I thought it would be the best to make Circle extending CircleMarker instead of having the same code twice. If you have time it would be nice, if can take a look into the code.

Together with @codeofsumit I will find another name then `editable` but else this PR should be ready to merge.

README.md
```diff
@@ -288,9 +288,10 @@ See the available options in the table below.
...
+| minRadiusCircleMarker | `null`                                | Set the min radius of a `CircleMarker`.                                                                                                                                                                         |
+| maxRadiusCircleMarker | `null`                                | Set the max radius of a `CircleMarker`.                                                                                                                                                                         |
+| editableCircle        | `true`                                | Enables radius editing while drawing a `Circle`.                                                                                                                                                                |
+| editableCircleMarker  | `false`                               | Enables radius editing while drawing a `CircleMarker`.                                                                                                                                                          |
```
codeofsumit, line 294: how about `resizableCircle` and `resizeableCircleMarker`

PR #1311, opened 2023-02-13
Enables rotation around any origin point

strfx:
Hi,

this PR extends `rotateLayer` and `rotateLayerToAngle` to take an optional point of rotation, allowing shapes to be rotated around *any* point in addition to the shape's center.

Example:

```javascript
// map setup [...] 

const coordinates = [
  {
    lat: 51.50795895196253,
    lng: -0.13981819152832034,
  },
  {
    lat: 51.518748822844444,
    lng: -0.13981819152832034,
  },
  {
    lat: 51.518748822844444,
    lng: -0.11990547180175783,
  },
  {
    lat: 51.50795895196253,
    lng: -0.11990547180175783,
  },
];

const poly = new L.polygon(coordinates, { color: 'red' }).addTo(map);

poly.pm.rotateLayer(45, coordinates[0]);
```

resulting in:

image

The point is optional. If no point is provided, it defaults to the current behaviour of rotating the shape around its center point. Please note that this only works when done programmatically, not interactively in rotation mode.

Let me know what you think!


strfx:
I've updated the PRs description with an example to better illustrate the change.

Also, do you generate the API docs in the README or update it manually? I could not find any hints on that.

Thanks! :-)

Falke-Design (changes requested):
I think it makes more sense to have a `setRotationCenter()` method and a public `getRotationCenter()`. With that the rotation point is stored on the layer and can be exported too.

Can you please also make a test to import a layer with GeoJSON and a different rotation point.

https://github.com/geoman-io/leaflet-geoman/blob/144c0c89a55a93f011d7b90170c7dee748de8189/src/js/Mixins/Rotating.js#L127

strfx:
good approach @Falke-Design , thanks! Will update the PR accordingly.

strfx:
> Can you please also make a test to import a layer with GeoJSON and a different rotation point.

@Falke-Design To clarify the testing requirement: were you thinking of testing a GeoJSON structure that contains information about the shape's rotation center (from in its `properties`), or just rotating any GeoJSON layer around a new rotation center?

P.S. I screwed up the PR by syncing the fork - I'll reopen a new PR.

PR #1322, opened 2023-02-27
Fix Norwegian translation typoes (lower case)

andreasvatne:
Every character should be lower case except for the first one in each sentence.

### DIFF
--- a/src/assets/translations/no.json
+++ b/src/assets/translations/no.json
@@ -16,15 +16,15 @@
     "removeLastVertex": "Fjern forrige punkt"
   },
   "buttonTitles": {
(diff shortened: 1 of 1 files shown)


Falke-Design:
Thank you, will be merged. Would it be possible for you to add the missing translations too?

https://github.com/geoman-io/leaflet-geoman/blob/ec9e647f83edbc25c1e4fe9336abf9a32d0f7c12/src/assets/translations/en.json

andreasvatne:
I've added Norwwegian translations for the remaining items as well.

codeofsumit (approved):

codeofsumit (approved):

PR #1343, opened 2023-04-11
Fix snapping removed layers

beig:
This small fix keeps the snapList updated if layers are removed other than with `RemovalMode`
also check for valid index when removing an item from the list


Falke-Design (approved):
Makes totally sense, thank you!

PR #1362, opened 2023-06-13
Set rotation center

strfx:
c.f. https://github.com/geoman-io/leaflet-geoman/pull/1311

This PR allowing shapes to be rotated around any point in addition to the shape's center.

### DIFF
--- a/cypress/integration/rotation.spec.js
+++ b/cypress/integration/rotation.spec.js
@@ -248,7 +248,7 @@ describe('Rotation', () => {
     });
   });
 
-  it('rotates around arbitrary origins', () => {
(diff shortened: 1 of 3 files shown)


Falke-Design (approved):
Very nice, thank you!

After merging, I need to update the docs:
```

      /** Returns the center of rotation. */
      getRotationCenter(): L.LatLng;

      /** Change the center of rotation. Pass null to use the shape's default center. */
      setRotationCenter(center: L.LatLng | null): void;
```

Falke-Design:
@strfx no, thank you. After a review of @codeofsumit we will merge it

PR #1372, opened 2023-06-24
Auto Prettier on commit

### LINKED ISSUE #1371: Add auto prettier to pre-commit hook
If a commit happens, the prettier should run automatically

midyh:
Closes #1371.

I added lint-staged and husky so that Prettier and ESLint are ran on commit.


package.json
```diff
@@ -86,5 +88,9 @@
     "tabWidth": 2,
     "semi": true,
     "singleQuote": true
+  },
+  "lint-staged": {
+    "*.js": "eslint --cache --fix",
+    "*.{js,css,md}": "prettier --write"
```
Falke-Design, line 94: Can you please apply the same rules as in the npm scripts. (I changed them just now a little bit)

```
"eslint": "eslint {src,cypress,demo}/**/*.js --fix ",
"prettier": "prettier --write {src,cypress,demo}/**/*.{js,css}",
```

Falke-Design (changes requested):
Looks already pretty good, thank you!

package.json
midyHamdoun: Done

Falke-Design (approved):

Issue #1369, opened 2023-06-24
Draw.Rectangle should use the same code to get the corners as Edit.Rectangle


Falke-Design:
The corners will be not correct when the rectangle is drawn with `rectangleAngle`.

- [ ] Update `_findCorners` in Draw.Rectangle
- [ ] add `rectangleAngle: 0` to the default Draw options: src/js/Draw/L.PM.Draw.js

Current:
https://github.com/geoman-io/leaflet-geoman/blob/932d4a986ec5853f183a1a69fa8d9d55bff963de/src/js/Draw/L.PM.Draw.Rectangle.js#L248-L257

New & correct one:
https://github.com/geoman-io/leaflet-geoman/blob/932d4a986ec5853f183a1a69fa8d9d55bff963de/src/js/Edit/L.PM.Edit.Rectangle.js#L199-L207

CYBWEBALI:
I can fix this issue. Assign it to me

CYBWEBALI:
Done PR #1373 .

PR #1373, opened 2023-06-27
fix: Draw.Rectangle should use the same code to get the corners as Edit.Rectangle

AlimurtuzaCodes:
### What does this PR do

Followed two steps as mentioned in the issue.

- [x] Update _findCorners in Draw.Rectangle
- [x] add rectangleAngle: 0 to the default Draw options: src/js/Draw/L.PM.Draw.js

Fixed Draw.Rectangle file in [leaflet-geoman/src/js/Draw/L.PM.Draw.Rectangle.js](https://github.com/geoman-io/leaflet-geoman/blob/932d4a986ec5853f183a1a69fa8d9d55bff963de/src/js/Draw/L.PM.Draw.Rectangle.js#L248-L257)

Issue Number- #1369

### DIFF
--- a/cypress/integration/rectangle.spec.js
+++ b/cypress/integration/rectangle.spec.js
@@ -880,4 +880,26 @@ describe('Draw Rectangle', () => {
       expect(layer.options.color).to.eql('red');
     });
   });
+
+  it('Return correct corners of rotated rectangle while drawing', () => {
+    cy.window().then(({ map }) => {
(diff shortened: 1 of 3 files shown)


Falke-Design:
Thank you for creating this PR. Do you think you are able to add a test for it too?

AlimurtuzaCodes:
> Thank you for creating this PR. Do you think you are able to add a test for it too?

I won't be able to add a test for this.

Falke-Design (approved):

Issue #1390, opened 2023-08-23
Disable finish drawing when existing vertex is clicked


AndrejGajdos:
When user is drawing a line and existing vertex is clicked, the drawing is finished. How can I disable this behaviour to finish drawing only when "Finish" button is clicked?

Falke-Design:
Currently this is not built in, but we will think about adding it.

You can use following code:

```
      const createMarkerProto = L.PM.Draw.Line.prototype._createMarker;
      const setTooltipTextProto = L.PM.Draw.Line.prototype._setTooltipText;
      L.PM.Draw.Line.include({
        _createMarker: function(latlng){
          const marker = createMarkerProto.call(this, latlng);
          marker.off('click', this._finishShape, this);
          return marker;
        },
        _setTooltipText: function() {
          setTooltipTextProto.call(this);
          const { length } = this._layer.getLatLngs().flat();
          if (length > 1) {
            this._hintMarker.setTooltipContent('Click finish in the Toolbar to create the Line');
          }
        },
      });
      map.pm.Draw.Line = new L.PM.Draw.Line(map);
```

AndrejGajdos:
@Falke-Design thank you, this is great improvement, but with this modification it's still possible to finish drawing when user clicks or double clicks the existing vertex.

![geoman-click-existing-vertex](https://github.com/geoman-io/leaflet-geoman/assets/1593758/e3b83e44-efb5-46aa-8d34-07a09663038e)

PR #1395, opened 2023-08-31
add finishOnVertexClick option

AndrejGajdos:
More context #1390


Falke-Design (changes requested):
Thank you for the PR.
I think it will be cleaner if we use `finishOn: 'finishBtn'` instead of the new option `finishOnVertexClick`.

Also please add some tests.

Issue #1422, opened 2023-11-22
Error when drawing text box after toolbar button is enabled and moving directly from mode other than draw


matthewtboyle:
Here is the JSFiddle https://jsfiddle.net/matthewtboyle/r1cdufga/3/

This fiddle is set up to disable or enable the drawText toolbar button on zoom end. When you open the fiddle:
1. click toolbar button to enter edit mode
2. zoom out so that the drawText button is disabled.
3. zoom in until drawText button is enabled
4. click the drawText button.

I don't see the details of the error in the JSFiddle, but in my environment it looks like this:

`global-error-handler.ts:22 handleError TypeError: Cannot read properties of undefined (reading 'remove')
    at NewClass.disable (leaflet-geoman.min.js:1:149230)
    at NewClass.toggle (leaflet-geoman.min.js:1:149529)
    at Object.afterClick (leaflet-geoman.min.js:1:80589)
    at NewClass._triggerClick (leaflet-geoman.min.js:1:69847)
    at NewClass.triggerClickOnToggledButtons (leaflet-geoman.min.js:1:76640)
    at NewClass.toggleButton (leaflet-geoman.min.js:1:76838)
    at NewClass.disableGlobalEditMode (leaflet-geoman.min.js:1:39953)
    at NewClass.toggleGlobalEditMode (leaflet-geoman.min.js:1:40328)
    at Object.afterClick (leaflet-geoman.min.js:1:79022)
    at NewClass._triggerClick (leaflet-geoman.min.js:1:69847)` 
    
It looks like it is trying to remove a hintMarker that does not exist.
    
I do know that when moving to draw mode from another active mode Geoman is disabling all other modes. This remove happens in L.PM.Draw.Text.disable(). Maybe enabling the drawText button is causing Geoman to think we need to run L.PM.Draw.Text.disable() when LPM.Draw.Text.toggle() is run?

Any help is appreciated.

Falke-Design:
Thank you for creating this issue. It will be fixed in the next version. #1424

Until the release you can use following workaround:
```
map.pm.Toolbar.buttons.drawText._button.onClick = function() {
  if (map.pm.Draw.Text._enabled && !map.pm.Draw.Text._hintMarker) {
    map.pm.Draw.Text._hintMarker = L.marker(map.getCenter());
  }
};
```
https://jsfiddle.net/pjshyoru/

PR #1424, opened 2023-11-22
After disabling & enabling of button, don't call disable on the draw layer. 

Falke-Design:
After disabling & enabling of button, while a mode is active, don't call disable on the draw layer.
And added that text disable is not throwing an error if hintMarker is not set.

Fixes: #1422

### DIFF
--- a/cypress/integration/toolbar.spec.js
+++ b/cypress/integration/toolbar.spec.js
@@ -501,7 +501,7 @@ describe('Testing the Toolbar', () => {
       });
   });
 
-  it("After disabling & enabling of button, while a mode is active, don't call disable on the draw layer", (done) => {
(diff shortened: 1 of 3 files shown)


cypress/integration/toolbar.spec.js
```diff
@@ -500,4 +500,24 @@ describe('Testing the Toolbar', () => {
         expect(eventFired).to.equal('drawPolygon');
       });
   });
+
+  it("After disabling & enabling of button, while a mode is active, don't call disable on the draw layer", (done) => {
```
Falke-Design, line 504:
```suggestion
  it("After disabling & enabling of a button, while a mode is active, don't call disable on the draw layer", (done) => {
```

PR #1425, opened 2023-11-26
[+] Add Kyrgyz language support

warioddly:


Falke-Design (changes requested):
Please also add the language code to the type definition:
https://github.com/geoman-io/leaflet-geoman/blob/b8b56e785deea901b94f2d8d683c70130f66327b/leaflet-geoman.d.ts#L498-L500

Falke-Design (approved):
Thank you for your contribution. It will be added in the next release but this can take a while.

PR #1434, opened 2024-01-02
fix global keyboard and window listeners are not removed after the map is destroyed

plainheart:
Hi,

I noticed the global listeners bound in the keyboard mixin were not unbound after the map was destroyed, which makes the listener always exist and makes the code throw errors even if the component is destroyed in those single-page applications.

So I think we should unbind the listeners somewhere after destroying the map. I currently put this logic into the `_initKeyListener` function of the `Keyboard` mixin. Also, to keep the isolation of the mixin, I changed the singleton Keyboard mixin object to a function that returns a new object at each call to make it work for multiple map instances, or the entire listener will be removed as long as one map is destroyed, which causes the event listener for other maps to not work.

To reproduce,

- https://codepen.io/plainheart/pen/bGZVxgv
- Or just open `demo/index.html` in this project and call `map.remove()`

### DIFF
--- a/cypress/integration/mixins.spec.js
+++ b/cypress/integration/mixins.spec.js
@@ -1,5 +1,5 @@
-describe('Should unbind event listeners that bound by the Mixins after the map is destroyed', () => {
-  it('KeyboardMixin', () => {
+describe('', (KeyboardMixin) => {
+  it('Should unbind event listeners that bound by the KeyboardMixin after the map is destroyed', () => {
     cy.window().then((window) => {
       const { map, document } = window;
 
--- a/cypress/integration/mixins.spec.js
+++ b/cypress/integration/mixins.spec.js
@@ -1,0 +1,25 @@
+describe('Should unbind event listeners that bound by the Mixins after the map is destroyed', () => {
+  it('KeyboardMixin', () => {
+    cy.window().then((window) => {
+      const { map, document } = window;
+
+      map.remove();
+
+      const isWindowBlurEventUnbound = !Object.entries(
+        window._leaflet_events
+      ).some(([name, handler]) => name.startsWith('blur') && handler);
+      expect(
+        isWindowBlurEventUnbound,
+        'window blur event listener is not unbound'
+      ).to.eq(true);
+
+      const isKeyUpDownEventUnbound = !Object.entries(
+        document._leaflet_events
+      ).some(([name, handler]) => name.startsWith('key') && handler);
+      expect(
+        isKeyUpDownEventUnbound,
+        'document keyboard event listener is not unbound'
+      ).to.eq(true);
+    });
+  });
+});
--- a/cypress/integration/mixins.spec.js
+++ b/cypress/integration/mixins.spec.js
@@ -1,4 +1,4 @@
-describe('', (KeyboardMixin) => {
+describe('KeyboardMixin', () => {
   it('Should unbind event listeners that bound by the KeyboardMixin after the map is destroyed', () => {
     cy.window().then((window) => {
       const { map, document } = window;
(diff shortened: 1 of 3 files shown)


Falke-Design (changes requested):
Thank you, good catch!

I will approve this but are you able to add some tests?

And do you think we have with some other Mixins problems when there are multiple maps?

plainheart:
@Falke-Design Just added a simple test case but I'm not sure if it's enough.

> And do you think we have with some other Mixins problems when there are multiple maps?

It looks other mixins are not binding events on the global objects like `window` and `document`. There might be no problem with them. One suggestion is to provide an API to allow users to unmount (or destroy/remove/dispose) the PM plugin and clean up the relative resources like listeners and mounted objects.

src/js/Mixins/Keyboard.js
```diff
@@ -1,9 +1,16 @@
...
   _initKeyListener(map) {
     this.map = map;
     L.DomEvent.on(document, 'keydown keyup', this._onKeyListener, this);
     L.DomEvent.on(window, 'blur', this._onBlur, this);
+    // clean up global listeners when current map instance is destroyed
+    map.on('unload', () => {
+      L.DomEvent.off(document, 'keydown keyup', this._onKeyListener, this);
+      L.DomEvent.off(window, 'blur', this._onBlur, this);
+    });
   },
```
Falke-Design, line 14: I think a named function would be better and cleaner. Also `once` should be enough
```suggestion
    map.once('unload', this_unbindKeyListenerEvents, this);
  },
  _unbindKeyListenerEvents() {
      L.DomEvent.off(document, 'keydown keyup', this._onKeyListener, this);
      L.DomEvent.off(window, 'blur', this._onBlur, this);
  },
```

cypress/integration/mixins.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('Should unbind event listeners that bound by the Mixins after the map is destroyed', () => {
+  it('KeyboardMixin', () => {
+    cy.window().then((window) => {
+      const { map, document } = window;
```
Falke-Design, line 4: This should be enough:
```suggestion
    cy.window().then(({ map, document }) => {
```

Falke-Design (changes requested):
This test is enough for now. Please check out the comments

cypress/integration/mixins.spec.js
Falke-Design: Never mind, i didn't see that you need the window object later

Falke-Design (approved):

cypress/integration/mixins.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('Should unbind event listeners that bound by the Mixins after the map is destroyed', () => {
+  it('KeyboardMixin', () => {
```
Falke-Design, line 2:
```suggestion
describe('', (KeyboardMixin) => {
  it('Should unbind event listeners that bound by the KeyboardMixin after the map is destroyed', () => {
```

src/js/Mixins/Keyboard.js
```diff
@@ -1,9 +1,17 @@
...
+// use function to create a new mixin object for keeping isolation
+// to make it work for multiple map instances
+const createKeyboardMixins = () => ({
   _lastEvents: { keydown: undefined, keyup: undefined, current: undefined },
   _initKeyListener(map) {
     this.map = map;
     L.DomEvent.on(document, 'keydown keyup', this._onKeyListener, this);
     L.DomEvent.on(window, 'blur', this._onBlur, this);
+    // clean up global listeners when current map instance is destroyed
+    map.once('unload', this_unbindKeyListenerEvents, this);
```
Falke-Design:
```suggestion
    map.once('unload', this._unbindKeyListenerEvents, this);
```

Falke-Design (changes requested):

cypress/integration/mixins.spec.js
Falke-Design: Sorry, I have copied to the wrong position

cypress/integration/mixins.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('', (KeyboardMixin) => {
```
Falke-Design, line 1:
```suggestion
describe('KeyboardMixin', () => {
```

Falke-Design (approved):
Thank you for your contribution!

PR #1443, opened 2024-01-31
Merge function

Rasovsky:


mscno:
This is a feature we are adding in the pro version, so we will not be adding this to the free version.

PR #1455, opened 2024-02-15
Minor: Force rotateEnabled() to always return a boolean.

strfx:
`_rotateEnabled` might not yet be initialized when calling `rotateEnabled`, returning `undefined` in that case:

https://github.com/geoman-io/leaflet-geoman/blob/7f388cc15ee5b7047bd997475ccaa7d30c7d15f0/src/js/Mixins/Rotating.js#L225

Hence converting to boolean first. Adopting this pattern from other places such as `layerDragEnabled`:

https://github.com/geoman-io/leaflet-geoman/blob/7f388cc15ee5b7047bd997475ccaa7d30c7d15f0/src/js/Mixins/Dragging.js#L133

### DIFF
--- a/cypress/integration/rotation.spec.js
+++ b/cypress/integration/rotation.spec.js
@@ -37,6 +37,7 @@ describe('Rotation', () => {
 
     cy.window().then(({ map }) => {
       const layer = map.pm.getGeomanDrawLayers()[0];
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):

Falke-Design:
Thank you very much!

PR #1456, opened 2024-02-16
Suppress errors in Snapping.js

tanthammar:
On mouse-over, markers, L.Divicons or sometimes layer groups, 
with a tool active, 
I get errors in console where this._snapList is undefined.

It originates from events dispatched from Leaflet when tooltips are opened/closed. 

This error sometimes breaks Geoman and I have to reload the page to continue.

A simple solution is to check the _snapList property.

This non-breaking change, solves the problem.

### DIFF
--- a/src/js/Mixins/Snapping.js
+++ b/src/js/Mixins/Snapping.js
@@ -258,7 +258,7 @@ const SnapMixin = {
     this.debugIndicatorLines = debugIndicatorLines;
   },
   _handleSnapLayerRemoval({ layer }) {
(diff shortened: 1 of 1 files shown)


tanthammar:
@Falke-Design I do understand that you want me to reproduce the error.
I use TALL-stack. Laravel, Alpine js, Livewire, Tailwind.
The task is to big for me to setup a demo.
I have made a fork where I added the change.
Thank you for replying and for maintaining this awesome tool!
Apologies for not going all the way with reproducing the error for you.

PR #1475, opened 2024-04-01
Allow removal of holes in polygons while maintaining the minimum vertex count of the geometry

### LINKED ISSUE #1474: Disabling "removeLayerBelowMinVertexCount" prevents removal of holes in polygon
This might be the intended behavior, but it seems odd.

When the “removeLayerBelowMinVertexCount” option is disabled, i.e., set to false, on a polygon with holes, you are unable to remove the holes in the polygon by right-clicking the vertices of the holes in edit mode.

Despite preventing the removal of the layer (i.e., the polygon itself), one would likely still expect to be able to remove parts of the layer (i.e., the holes themselves).

https://jsfiddle.net/kas04bf3/7/

Tobikblom:
Fixes #1474.

Also, please verify that `parentPath[parentPath.length - 1] > 0;` checks if marker is part of a hole.

### DIFF
--- a/src/js/Edit/L.PM.Edit.Line.js
+++ b/src/js/Edit/L.PM.Edit.Line.js
@@ -473,8 +473,11 @@ Edit.Line = Edit.extend({
     let markerArr =
       indexPath.length > 1 ? get(this._markers, parentPath) : this._markers;
 
-    // prevent removal of the layer if the vertex count is below minimum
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Thank you for submitting this PR. Can you please add a test as well?

Falke-Design (approved):

PR #1499, opened 2024-06-13
fixes setLang to handle navigator.language formats (such as "fr-FR")

### LINKED ISSUE #1472: setLang works with "fr" but not "fr-FR"
I use `navigator.language` which can return values likes "en", "en-US", "fr", "fr-FR", "es-ES", etc.

It seems calling `setLang` using the 2 letter value like "fr" works, but not the 5 letter value like "fr-FR". There should be a fallback from "fr-FR" to "fr", "es-ES" to "es", ...Basically just keeping the 1st 2 letters should fix it.

mscno:
closes #1472

### DIFF
--- a/src/assets/translations/index.js
+++ b/src/assets/translations/index.js
@@ -29,6 +29,9 @@ import fi from './fi.json';
 import ko from './ko.json';
 import ky from './ky.json';
 
+// eslint-disable-next-line camelcase
+const pt = pt_pt;
+
 export default {
   en,
   de,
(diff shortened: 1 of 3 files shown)


src/assets/translations/index.js
```diff
@@ -39,6 +42,7 @@ export default {
   es,
   nl,
   fr,
+  pt,
```
Falke-Design, line 45: we should add this in the typescript file too

Falke-Design (changes requested):

mscno (commented):
great!

Falke-Design (approved):

PR #1514, opened 2024-09-15
Fix missing type definition

m208:
Fixed missing type definition for the "name" property of the Action object, which is used as the class name for the button.  

Source:
https://github.com/geoman-io/leaflet-geoman/blob/9f6aac717f2909fbc12467b7e1dca6ec091b2fd3/src/js/Toolbar/L.Controls.js#L163-L176   

TS error:

image


Falke-Design (approved):
Thank you for fixing this.

Issue #1111, opened 2022-02-18
TypeScript EventHandlers do not support all properties in payload


Falke-Design:
In the event payload there is always the `source` and the possibility to add custom properties. Is it possible to extend types and make a base event type with `source` and custom properties?

TS:
https://github.com/geoman-io/leaflet-geoman/blob/1fdc918fa39ffa84327fdf639fa75865168f716d/leaflet-geoman.d.ts#L1437-L1440

JS where the payload and the custom payload are merged:
https://github.com/geoman-io/leaflet-geoman/blob/1fdc918fa39ffa84327fdf639fa75865168f716d/src/js/Mixins/Events.js#L615-L618

@ryan-morris maybe you have time to check this?

PR #1523, opened 2024-10-13
TypeScript fix custom props #1111 

DoiMayank:
#1111 
Extend event handlers to support custom properties in TypeScript

- Created a BaseEventType with source and custom properties.
- Extended event-specific types from the base type to ensure consistency.
- Updated the _fire method in Events.js to merge custom payload with source.

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -1776,10 +1776,15 @@ declare module 'leaflet' {
     /**
      * ROTATE MODE MAP EVENT HANDLERS
      */
-    export type GlobalRotateModeToggledEventHandler = (e: {
+    export type BaseEventHandler = (e: { 
+      source: any; 
+      custom?: Record<string, any>; 
+   }) => void;
+
+    export type GlobalRotateModeToggledEventHandler = BaseEventHandler & {
       enabled: boolean;
       map: L.Map;
-    }) => void;
+  };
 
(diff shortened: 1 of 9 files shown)


src/js/Mixins/Events.js
```diff
@@ -698,7 +698,7 @@ const EventMixin = {
 
   // private (very private) fire function
   __fire(fireLayer, type, payload, source, customPayload = {}) {
-    payload = merge(payload, customPayload, { source });
```
Falke-Design, line 701: Why is this needed?

leaflet-geoman.d.ts
```diff
@@ -1776,10 +1776,15 @@ declare module 'leaflet' {
     /**
      * ROTATE MODE MAP EVENT HANDLERS
      */
-    export type GlobalRotateModeToggledEventHandler = (e: {
+    export type BaseEventHandler = (e: { 
+      source: any; 
+      custom?: Record<string, any>; 
+   }) => void;
+
+    export type GlobalRotateModeToggledEventHandler = BaseEventHandler & {
```
Falke-Design, line 1784: All the other Events need the BaseEventHandler too.
Additionally I think we should provide a GlobalModeBaseEventHandler, because most of them are the same

Falke-Design (changes requested):

PR #1569, opened 2025-03-12
Prevent the creation of an additional Marker when dragging a Marker i…

### LINKED ISSUE #1568: Bug: Dragging a Marker while creating Markers results in a new Marker being created at the end of the drag
### Web Browser (with version)

chrome, firefox

### OS/Platform (with version)

Windows

### Leaflet version

1.9.4

### Geoman version

2.18.3

### Geoman Pro

- [ ] I'm using Geoman Pro

### Minimal example reproducing the issue

demo/index.html

### Steps to reproduce

1. Open leaflet-geoman/demo/index.html
2. Enter Marker Draw mode by clicking the "Draw Marker" button on the Control bar
3. Create a Marker by clicking on the map
4. Drag the Marker by clicking and dragging the Marker
5. When you release the mouse, the Marker that was dragged will be position appropriately.  However, an additional Marker will be created where the cursor marker is
6. Due to snapping, the Marker can be created at the same position.  Using the Remove Layers tool, you can still tell there are 2 Markers.

### Expected behavior

When dragging a Marker, the Marker should be dragged without creating an additional Marker.

NOTE: This works for as expected for CircleMarker.

### Current behavior

Dragging a Marker while in "Draw Marker" mode creates an additional Marker.  CircleMarkers behave correctly.

tfitzsim:
Prevent the creation of an additional Marker when dragging a Marker in "Create Marker" mode. Fixes: #1568 

  - Replicate CircleMarker _layerIsDragging functionality in Marker

### DIFF
--- a/cypress/e2e/marker.cy.js
+++ b/cypress/e2e/marker.cy.js
@@ -317,27 +317,31 @@ describe('Draw Marker', () => {
     });
   });
 
-  it('drag in draw mode does not create additional marker', (done) => {
+  it('does not create additional marker while dragging in draw mode ', () => {
     cy.toolbarButton('marker').click();
 
     cy.get(mapSelector).click(150, 250);
-    cy.wait(1000);
 
     cy.window().then(({ map }) => {
       expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
     });
 
-    cy.get(mapSelector).trigger('mousemove', 150, 230);
-    cy.get(mapSelector).trigger('mouseover', 150, 230, { which: 1 });
     cy.get(mapSelector).trigger('mousedown', 150, 230, { which: 1 });
     cy.get(mapSelector).trigger('mousemove', 170, 290, { which: 1 });
+    // Do not create a new marker while dragging
     cy.get(mapSelector).click(170, 290);
-    cy.get(mapSelector).trigger('mouseup', 170, 290, { which: 1, force: true });
+    cy.get(mapSelector).trigger('mouseup', 170, 290, { which: 1 });
     cy.get(mapSelector).trigger('mousemove', 190, 340, { which: 1 });
 
     cy.window().then(({ map }) => {
       expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
-      done();
+    });
+
+    // Create a new marker after dragging with clicking on the icon of a marker
+    cy.get(mapSelector).click(170, 290);
+
+    cy.window().then(({ map }) => {
+      expect(map.pm.getGeomanDrawLayers().length).to.eq(2);
     });
   });
 });
--- a/cypress/e2e/marker.cy.js
+++ b/cypress/e2e/marker.cy.js
@@ -316,4 +316,28 @@ describe('Draw Marker', () => {
       expect(layer._icon.src.endsWith('someIcon.png')).to.eql(true);
     });
   });
+
+  it('drag in draw mode does not create additional marker', (done) => {
+    cy.toolbarButton('marker').click();
+
+    cy.get(mapSelector).click(150, 250);
+    cy.wait(1000);
+
+    cy.window().then(({ map }) => {
+      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
+    });
+
+    cy.get(mapSelector).trigger('mousemove', 150, 230);
+    cy.get(mapSelector).trigger('mouseover', 150, 230, { which: 1 });
+    cy.get(mapSelector).trigger('mousedown', 150, 230, { which: 1 });
+    cy.get(mapSelector).trigger('mousemove', 170, 290, { which: 1 });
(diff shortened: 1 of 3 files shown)


Falke-Design (changes requested):
Looks already good, thank you. Can you please add a test as well?

tfitzsim:
> Looks already good, thank you. Can you please add a test as well?

I appear to be out of my depth, so I'll need some help with the test.  I tried to get it working for quite a while, but couldn't.

1. I can create a Marker with click(), but not with trigger(mousedown/mouseup).
2. I could drag the Marker with both trigger(mousedown/mousemove/mouseup) and "touchMarker" logic used in other tests
3. However, after the Marker was dragged and the mouse moved away, there was still only 1 Marker
4. The issue was easy to trigger manually after the test completed.

I doubt this will be overly useful, but he is the basis of the code I was trying:
```
  it('drag in draw mode does not create additional marker', (done) => {
    cy.toolbarButton('marker').click();
    cy.get(mapSelector).click(150, 250);
    cy.wait(1000);

    cy.window().then(({ map }) => {
      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
    });

    cy.get(mapSelector).trigger("mousemove", 150, 230);
    cy.wait(1000);
    cy.get(mapSelector).trigger("mouseover", 150, 230, { which: 1 });
    cy.wait(1000);
    cy.get(mapSelector).trigger("mousedown", 150, 230, { which: 1 });
    cy.wait(1000);
    cy.get(mapSelector).trigger("mousemove", 170, 290, { which: 1 });
    cy.wait(1000);
    cy.get(mapSelector).trigger("mouseup", 170, 290, { which: 1, force: true });
    cy.wait(1000);
    cy.get(mapSelector).trigger("mousemove", 190, 340, { which: 1 });
    cy.wait(1000);

    cy.window().then(({ map }) => {
      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
      done();
    });
  });

```

tfitzsim:
PR is updated with a test.  I believe simulating a click before the mouseup is the correct thing to do.  The test fails without the fix, and succeeds with it.

cypress/e2e/marker.cy.js
```diff
@@ -316,4 +316,28 @@ describe('Draw Marker', () => {
...
+    cy.get(mapSelector).trigger('mousemove', 170, 290, { which: 1 });
+    cy.get(mapSelector).click(170, 290);
+    cy.get(mapSelector).trigger('mouseup', 170, 290, { which: 1, force: true });
+    cy.get(mapSelector).trigger('mousemove', 190, 340, { which: 1 });
+
+    cy.window().then(({ map }) => {
+      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
+      done();
+    });
+  });
```
Falke-Design:
```suggestion
  it('does not create additional marker while dragging in draw mode ', () => {
    cy.toolbarButton('marker').click();

    cy.get(mapSelector).click(150, 250);

    cy.window().then(({ map }) => {
      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
    });

    cy.get(mapSelector).trigger('mousedown', 150, 230, { which: 1 });
    cy.get(mapSelector).trigger('mousemove', 170, 290, { which: 1 });
    // Do not create a new marker while dragging
    cy.get(mapSelector).click(170, 290);
    cy.get(mapSelector).trigger('mouseup', 170, 290, { which: 1 });
    cy.get(mapSelector).trigger('mousemove', 190, 340, { which: 1 });

    cy.window().then(({ map }) => {
      expect(map.pm.getGeomanDrawLayers().length).to.eq(1);
    });

    // Create a new marker after dragging with clicking on the icon of a marker
    cy.get(mapSelector).click(170, 290);

    cy.window().then(({ map }) => {
      expect(map.pm.getGeomanDrawLayers().length).to.eq(2);
    });
  });
```

Falke-Design (commented):
Yes normaly the `click` events comes after `mouseup` but in the simulation we are to slow and dragging would be already disabled. So we need to call `click` before `mouseup`.

grafik

Do not add `cy.wait()` if not needed. Everything else looked good. I changed the test a little bit to check if marker creation is working after dragging is disabled.

cypress/e2e/marker.cy.js
```diff
@@ -316,4 +316,32 @@ describe('Draw Marker', () => {
       expect(layer._icon.src.endsWith('someIcon.png')).to.eql(true);
     });
   });
+
+  it('does not create additional marker while dragging in draw mode ', () => {
```
Falke-Design:
```suggestion
  it('does not create additional marker while dragging in draw mode', () => {
```

Falke-Design (approved):

Falke-Design:
Thank you for your contribution. If you want you can add the same test for CircleMarker.

Issue #1407, opened 2023-10-26
Adding createCustomControl get Error: Button with this name already exists


Bahaa-Fadhil:
I try to add **Toolbar.createCustomControl** but get error: " **Button with this name already exists** "

See the code below for more details:

```
export const LeafletMapToolbarCustomControl = memo(() => {
    const map = useMapStore(state => state.map)
    const isToolbar = useLeafletMapIsToolbarStore(state => state.isToolbar)


    useEffect(() => {
        if (!map) return

        // add leaflet.pm controls to the map
        map.pm.addControls({
            drawControls: true,
            editControls: false,
        })

        const customButton: L.PM.CustomControlOptions = {
            name: "custom",
            title: "Custom",
            className: "control-icon-leaflet-map",
            onClick: () => {
                // Custom action when clicking the control
                // eslint-disable-next-line no-alert
                alert("Custom Control Clicked 🙋‍♂️")
            },
            toggle: true,
            block: "custom",
            actions: ["finishMode"],
        }

        // Add or remove the control based on the isMapTools state.
        if (isToolbar) {
         // Adding the button
            map.pm.Toolbar.createCustomControl(customButton)

            map.pm.addControls({
                custom: true,
                position: "topright",
                oneBlock: true,
            })
        } else {
            //   leafletMap.removeControl(control)
            // map.pm.removeControls()
        }

        // Clean up the control when the component unmounts
        // eslint-disable-next-line consistent-return
        return () => {
            map.pm.removeControls()
        }
    }, [map, isToolbar])

    return <div></div>
})

```

I am using package: 

 "leaflet": "^1.9.4",
 "@geoman-io/leaflet-geoman-free": "^2.14.2", (free version).
 
 **Is there any solution for this problem ?**

Falke-Design:
I think your problem is, that it will be rendered multiple times. Make sure that your code is called only once

DedaDev:
I'm having the same issue using React 19.
It seems like `map.pm.removeControls()` does not remove buttons created with `map.pm.Toolbar.createCustomControl`. I'll try to dive deeper into this issue.

PR #1575, opened 2025-03-29
unmounting custom controls

DedaDev:
This PR is addressing #1407 and [this](https://stackoverflow.com/questions/73737137/leaflet-geoman-button-with-this-name-already-exists-error-when-creating-a-new) issue

Creating a custom control inside `useEffect` would throw the error 'Button with this name already exists'. This occurs because there was no option to remove the custom button from the toolbar when a component unmounts.

[codesandbox](https://codesandbox.io/p/devbox/react-leaflet-bug-forked-xcmc8y?file=%2Fsrc%2FApp.js&workspaceId=ws_QDWYkU1Sa8Wp6ooBegsLqN)

This is a crucial feature for the React app as it removes custom components from both the toolbar and the UI. Currently, we only have the `removeControls` function, which removes controls from the UI.

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -1088,7 +1088,7 @@ declare module 'leaflet' {
       /** Disable button by control name */
       setButtonDisabled(name: TOOLBAR_CONTROL_ORDER, state: boolean): void;
 
-      /** Removes a custom Control from the Toolbar */
+      /** Deletes and removes a Control from the Toolbar */
       deleteControl(name: string): void;
     }
 
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -1087,6 +1087,9 @@ declare module 'leaflet' {
 
       /** Disable button by control name */
       setButtonDisabled(name: TOOLBAR_CONTROL_ORDER, state: boolean): void;
+
+      /** Removes a custom Control from the Toolbar */
+      deleteControl(name: string): void;
     }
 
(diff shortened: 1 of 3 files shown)


Falke-Design (changes requested):
Let us add in `L.PM.Toolbar` the new method `deleteControl` and make it only callable there. The custom control creation is called over `map.pm.Toolbar` too. 

We don't need a option to delete all controls at once. The user can keep track of the custom controls by him self.

Please revert your changes and add the following code in Toolbar and adjust the type definition.
```
  deleteControl(name) {
    const btnName = this._btnNameMapping(name);
    if (this.buttons[btnName]) {
      this.buttons[btnName].remove();
      delete this.buttons[btnName];
    }
  }
```

DedaDev:
> The user can keep track of the custom controls by him self.

Yeah, that is actually a better approach. 
Updated 👍

leaflet-geoman.d.ts
```diff
@@ -1087,6 +1087,9 @@ declare module 'leaflet' {
 
       /** Disable button by control name */
       setButtonDisabled(name: TOOLBAR_CONTROL_ORDER, state: boolean): void;
+
+      /** Removes a custom Control from the Toolbar */
```
Falke-Design:
```suggestion
      /** Deletes and removes a Control from the Toolbar */
```

Falke-Design (changes requested):

Falke-Design (approved):

PR #1577, opened 2025-04-04
Prevent malformed HTML resulting from SVG file

JonathanWbn:
There is an issue when bundling the `@geoman-io/leaflet-geoman-free/dist/leaflet-geoman.css` file with Vite.

What happens is that the `<style>` tag of the `Text.svg` file does not get properly escaped and thus produces malformed HTML.

This PR moves the style rules of this SVG from an inline `<style>` tag to SVG HTML properties.


iKasch (approved):

iKasch:
Nice one!

Falke-Design (approved):
Thank you!

Issue #1595, opened 2025-10-10
Feature Request: Better tooltip hints for the marker tool on touch screens


ahukkanen:
### Description

When testing this library on touch devices, we have noticed that the marker tool is rather confusing on touch screens as they do not have a cursor on the screen. The marker tool creates a hint marker that follows the mouse movements and on desktop with a "fine" pointer device (i.e. a mouse) this works great. This is the code that creates the hint marker (and there's also some other code in that class dealing with placing that marker):

https://github.com/geoman-io/leaflet-geoman/blob/d7203056957fc61da1824c91c845a7ea472e636b/src/js/Draw/L.PM.Draw.Marker.js#L29-L33

However, on touch devices we found this confusing because of few reasons:

- The tooltip attached to the hint marker suggests clicking it to place the marker on the map (i.e. click the hint marker, not where you actually want the new marker to be placed)
- Sometimes the hint marker hides the actually added marker which can cause confusion (was it added or not?)
- Sometimes it feels that the marker was not added to the map as nothing seems to happen and there is no visible feedback for the user

Here is an example of the current user experience:

https://github.com/user-attachments/assets/c1f2967e-29ea-43a1-bae7-b501828b49af

For touch screens I would rather suggest to display some kind of a "sticky" note that would tell the user what to do and hide the current hint marker.

To explain it more visually, here is an example of how this could work:

https://github.com/user-attachments/assets/1c944cb3-61bc-4914-8b98-1ffcaff4e2f4

There is some discussion over the internet that such features that depend on detecting whether the device has touch capabilities are not reliable and instead, the functionality should be the same whether you have a pointer device available or not. But for this particular case, I feel this would be an improvement for the users who do not have a pointing device, as the current behavior somewhat assumes that there is a fine pointing device available (as it listens to the `mousemove` events).

This is the JS code that I would use to detect if the user has a pointing device:

```js
window.matchMedia("(pointer:fine)").matches
```

When this is `true`, the user has a "fine" pointing device, which would typically be a mouse. I believe this is not too complex code and also fairly widely supported by now.

Thank you for this awesome library!

Falke-Design:
Hi,

thank you for your feedback. That is a nice idea, it would be nice if you could create a PR for that.

For checking mobile devices please use: `L.Browser.mobile`

And let us use a global option on the map to enable this "mobile" mode.

ahukkanen:
Thank you for your response! Great to hear that this would be something to consider.

[...]

PR #1597, opened 2025-10-16
Add sticky marker tooltip for touch devices

JoonasAapro:
This PR is created to add a feature requested in the issue [#1595](https://github.com/geoman-io/leaflet-geoman/issues/1595), which adds a sticky tooltip for the markers in case of the user not having a pointer device (e.g. a tablet or a phone).

It was confusing for touch screens to place the marker since it didn't have any feedback and sometimes the hint marker hides the actually added marker which can cause confusion.

More data and use case -examples in video form you can find by clicking the issue number.

### DIFF
--- a/src/js/Draw/L.PM.Draw.Marker.js
+++ b/src/js/Draw/L.PM.Draw.Marker.js
@@ -1,5 +1,5 @@
 import Draw from './L.PM.Draw';
-import { getTranslation } from '../helpers';
+import { getTranslation, deviceHasFinePointer } from '../helpers';
 
 Draw.Marker = Draw.extend({
   initialize(map) {
@@ -62,6 +62,8 @@ Draw.Marker = Draw.extend({
         }
       });
     }
+
+    this._setupTouchScreenTips();
 
     // fire drawstart event
     this._fireDrawStart();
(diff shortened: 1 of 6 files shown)


src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
       this._hintMarker?.setIcon(this.options.markerStyle.icon);
     }
   },
+  _setupTouchScreenTips() {
+    let markerHintTip = null;
```
Falke-Design: Please add the variable to the context: `this._markerHintTip`

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
...
+
+    this._map.on('pm:drawstart', (event) => {
+      if (event.shape === 'Marker' && !deviceHasFinePointer()) {
+        // Hides the "hint marker" as that is confusing for touch screen users.
+        const drawMarker = this._map.pm.Draw.Marker;
```
Falke-Design: creating the variable `drawMarker` is obsolete

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
...
+        // Creates the touch screen hint "permanently" (while using the tool)
+        // sticked at the top of the map.
+        markerHintTip = L.tooltip([0, 0], {
+          className: 'leaflet-tooltip-stickynote',
+          content: getTranslation('tooltips.placeMarkerTip'),
+          direction: 'bottom',
+          permanent: true,
+        }).addTo(this._map);
+
+        const updateHandler = () => this._updateHintPosition(markerHintTip);
```
Falke-Design: The call should be throttled `this._markerHintTipUpdateHandler = L.Util.throttle(`

Reference: https://github.com/geoman-io/leaflet-geoman/blob/d7203056957fc61da1824c91c845a7ea472e636b/src/js/Mixins/Modes/Mode.Drag.js#L15-L21

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
...
+        this._map.on('zoomlevelschange resize move', updateHandler, this);
+      }
+    });
+    this._map.on('pm:drawend', (event) => {
+      if (event.shape === 'Marker' && markerHintTip) {
+        markerHintTip.remove();
+        markerHintTip = null;
+        this._map.off(
+          'zoomlevelschange resize move',
+          this._updateHintPosition,
```
Falke-Design: This doesn't remove the correct event listener function

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
...
+          this
+        );
+      }
+    });
+  },
+  _updateHintPosition(markerHintTip) {
```
Falke-Design: No need to have `markerHintTip` as argument, if assigned to the context `this._markerHintTip`

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
...
+
+    this._map.on('pm:drawstart', (event) => {
+      if (event.shape === 'Marker' && !deviceHasFinePointer()) {
+        // Hides the "hint marker" as that is confusing for touch screen users.
+        const drawMarker = this._map.pm.Draw.Marker;
+        drawMarker._hintMarker.remove();
+
+        // Creates the touch screen hint "permanently" (while using the tool)
+        // sticked at the top of the map.
+        markerHintTip = L.tooltip([0, 0], {
```
Falke-Design: Not sure if we should use a plain html element, positioned absolute, instead. Then there would be no need to recalculate the position with every move

src/js/Draw/L.PM.Draw.Marker.js
```diff
@@ -194,4 +196,62 @@ Draw.Marker = Draw.extend({
       this._hintMarker?.setIcon(this.options.markerStyle.icon);
     }
   },
+  _setupTouchScreenTips() {
+    let markerHintTip = null;
+
+    this._map.on('pm:drawstart', (event) => {
```
Falke-Design: Eventlistener should be only added if the global option for mobile tooltip is enabled

Falke-Design (changes requested):
Looks already pretty good

JoonasAapro:
@Falke-Design Sorry for the delay regarding the changes!

- Joonas

mscno:
Hi @JoonasAapro, appreciate the work here. I took inspiration from your changes and added support for touch hints in a slightly different way. Added in PR https://github.com/geoman-io/leaflet-geoman/pull/1627

PR #1600, opened 2025-10-27
bump turf to 7x

mscno:
This PR updates Turf.js libraries to v7.2.0

### DIFF
--- a/cypress/e2e/polygon.cy.js
+++ b/cypress/e2e/polygon.cy.js
@@ -703,7 +703,8 @@ describe('Draw & Edit Poly', () => {
         const handSelfIntersectionFalse = new Hand({
           timing: 'frame',
           onStop() {
-            expect(poly.pm.hasSelfIntersection()).to.equal(false);
+            // commenting out this line, fixes the test?? Why do we expect this assertion to be false here? 
+            //expect(poly.pm.hasSelfIntersection()).to.equal(false);
 
             // Map shouldn't be dragged
             const center = map.getCenter();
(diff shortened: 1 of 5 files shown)


mscno:
@Falke-Design im trying to upgrade the turs.js libraries to use 7.x.x versions. This breaks a selfintersection tests in the e2e/polygons.spec.js test file. Can you have a look at this to see if we can fix this?

mscno:
I think it has to do with https://github.com/Turfjs/turf/pull/2772, https://github.com/Turfjs/turf/issues/2722, https://github.com/Turfjs/turf/discussions/2728

cypress/e2e/polygon.cy.js
```diff
@@ -703,7 +703,8 @@ describe('Draw & Edit Poly', () => {
         const handSelfIntersectionFalse = new Hand({
           timing: 'frame',
           onStop() {
-            expect(poly.pm.hasSelfIntersection()).to.equal(false);
+            // commenting out this line, fixes the test?? Why do we expect this assertion to be false here? 
+            //expect(poly.pm.hasSelfIntersection()).to.equal(false);
```
mscno: dropping the assertions here fixes the test? why are we expecting the polygon selfIntersection check to be false here?

i guess it should be true, since we havent modified the fixture at this point?

Falke-Design: just dropping something is never good 😄 

The fixture is moved by this lines https://github.com/geoman-io/leaflet-geoman/pull/1600/files#diff-50ef12e5ce5f469968d02a4a6da456f860b1731c5a119165e57e39a5eb9a184aR749-R769 and after that `onStop` is called. The selfIntersection needs to be resolved, that is whole point of this test 😄

Falke-Design:
@mscno fixed

PR #1604, opened 2025-11-16
Fix event listener for hint marker display

### LINKED ISSUE #1603: _showHintMarker is undefined
Should this be `_showHintMarkerAfterMoving`?
https://github.com/geoman-io/leaflet-geoman/blob/2fe81d8b371a7349f6fd131a4725651935cca785/src/js/Draw/L.PM.Draw.Text.js#L84

Schlepptop:
Fixes #1603


Falke-Design:
Thank you

PR #1646, opened 2026-04-09
fix(types): add source and custom properties to event handler types

### LINKED ISSUE #1111: TypeScript EventHandlers do not support all properties in payload

mvanhorn:
## Summary

All 73 PM event handler type definitions now include `source` and support custom payload properties, matching the runtime behavior in `Events.js`.

## Why this matters

The `__fire` method in `Events.js:700` merges `{ source }` and `customPayload` into every event payload. But the TypeScript types didn't reflect this, so `e.source` and any custom properties were untyped.

## Changes

- `leaflet-geoman.d.ts`: Added `BaseEventPayload` interface with `source: string` and `[key: string]: any`. Updated all 73 event handler types to use `BaseEventPayload & { ... }` intersection.

Fixes #1111

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/leaflet-geoman.d.ts
+++ b/leaflet-geoman.d.ts
@@ -2289,19 +2289,27 @@ declare module 'leaflet' {
      * DRAW MODE MAP EVENT HANDLERS
      */
 
-    export type GlobalDrawModeToggledEventHandler = (event: {
+    /** Base properties present on all PM event payloads. */
+    export interface BaseEventPayload {
+      /** The source that triggered the event (e.g. 'Draw', 'Edit', 'Global'). */
+      source: string;
+      /** Custom payload properties merged at fire time. */
+      [key: string]: any;
+    }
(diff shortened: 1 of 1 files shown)


Falke-Design:
Thank you for your help

PR #1651, opened 2026-04-17
docs: document TypeScript casting of event payload

mvanhorn:
## Description

Adds a short `### TypeScript` subsection to `README.md` (directly under `## Documentation`) showing how to narrow the `layer` property on Geoman event payloads from `L.Layer` to a shape-specific type. Two patterns are documented:

1. `instanceof` narrowing (recommended) — TypeScript narrows automatically inside the `if` block, no cast needed.
2. `as` cast branched on `e.shape` — for users who already know the shape via the `PM.SUPPORTED_SHAPES` enum.

The second snippet is the exact pattern requested in #1014. The event payload type is `PM.EditEventHandler` (defined in `leaflet-geoman.d.ts`), referenced in the docs so users can find the source of truth.

## Related Issue

Fixes #1014

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)

## Checklist

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) guidelines
- [x] My code follows the code style of this project (docs-only change; `npm run lint` unaffected)
- [x] I have tested my changes locally (rendered README preview; markdown renders cleanly)
- [x] I have added tests that prove my fix is effective or that my feature works (N/A — docs only)
- [x] New and existing tests pass locally with my changes (docs-only; no code paths touched)

## Screenshots (if applicable)

N/A — markdown-only change.

## Additional Notes

The issue was open for ~4.5 years (filed 2021-10-19) and references Stack Overflow question 69621270 as additional context for users hitting this. The example from the issue body is preserved verbatim in the second code block.

Fixes #1014

This contribution was developed with AI assistance (Codex).

PR #1656, opened 2026-05-04
docs(readme): document TypeScript casting of Geoman event payload

### LINKED ISSUE #1014: Document Typescript casting of event payload
In the event payload the return type of `layer` is `L.Layer? so it needs to be casted to the Layer-Type.

```
layer.on('pm:edit', (e) => {
  if(e.shape === 'Polygon'){
     (e.layer as Polygon).getLatLngs();
  }
});
```

This should documented in the Readme

Explainations:
- https://stackoverflow.com/questions/69621270/how-to-get-latlngs-values-after-edit-a-polygon-in-leaflet-geoman/69626304#69626304
- https://github.com/geoman-io/leaflet-geoman/issues/945

mvanhorn:
Closes #1014.

Geoman event payloads type `layer` as `L.Layer`, so reaching for layer-specific methods (e.g. `getLatLngs()` on a `Polygon`) requires casting or narrowing first. The README didn't show that pattern, so users hit the same TypeScript error the issue reporter ran into (also surfaced on [stackoverflow.com/q/69621270](https://stackoverflow.com/q/69621270)).

Adds a **TypeScript: casting the event payload** subsection under `## Documentation` that:

- Branches on `e.shape` and casts `e.layer` to the concrete subtype - the original snippet from the issue.
- Shows the equivalent `instanceof` narrowing form (which the issue reporter floated as untested but cleaner; it does work and reads better in modern TS).
- Links back to #1014 for the original discussion so future readers can find the rationale.

External `geoman.io/docs` is the deeper reference, but the README is what most TypeScript users hit first when they wire up `pm:edit` and the casting pattern is short enough to live there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,29 @@
 
 Visit [geoman.io/docs](https://www.geoman.io/docs) to get started.
 
+### TypeScript: casting the event payload
+
+The `layer` field on Geoman event payloads is typed as `L.Layer`, so reach for the concrete subtype before calling layer-specific methods. Branch on `e.shape` and cast, or narrow with `instanceof` first:
+
+```typescript
+import { Polygon } from 'leaflet';
+
+map.on('pm:edit', (e) => {
+  if (e.shape === 'Polygon') {
+    (e.layer as Polygon).getLatLngs();
+  }
+});
+
+// or, equivalently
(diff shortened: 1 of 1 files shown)


Falke-Design:
Still the same answer as in #1651

